Issue #47, opened 2018-02-14
SynchronousExecutor and stack overflow


davidmoten:
`DefaultExecutors` offers `performOnForeground` overloads. These overloads use a `SynchronousExecutor` under the covers. More than just being a synchronous executor it is an immediate executor so that recursive scheduling can cause stack overflow. A more robust implementation of this would be a trampoline scheduler that places new tasks on a queue, polls the queue for tasks and runs them (all synchronously).

Happy to contribute if desired.

deepak2016:
@davidmoten wanted to understand more on what do you mean by recursive scheduling which can cause stackoverflow error. I looked into SynchronousExecutor class and though comments in the class mention that this executor executes task in main thread, I don't see any mechanism using which this SynchronousExecutor will be able to execute task in main thread. 
This class Looks more like it will execute anything it is given, in its current thread only. There is no switching or mechanism to dispatching to main thread.

davidmoten:
@deepak2016 Look at the unit test `testManyReentrantCalls` in #71

deepak2016:
I went through the pull request and unit test and here is my understanding:  if everything is happening on same thread, even if you are putting tasks in a queue, it is synchronously executing the task in that thread. In unit test it seems to recursively calling same method and since everything is synchronous, it will still throw stackoverflow exception eventually.

deepak2016:
@davidmoten I went through the pull request and seems this will be useful for scenarios where someone is executing recursively on SynchronousExecutor. I am thinking where can this scenario arise in the SDK. This is an internal class which gets used by request classes. Can you provide any scenario which can cause recursive execution of SynchronousExecutor?

davidmoten:
@deepak2016 good question, but I don't agree that its an internal class at the moment. When you build a client `GraphServiceClient` the user can specify his/her own executors. The `DefaultExecutors` class includes a foreground executor which uses the `SynchronousExecutor` class. The internal usage in requests is limited to using IExecutors instances without knowledge of what instances they are. 

Ok so if a user uses the foreground executor of the DefaultExecutors class in a particular way we have established that without PR #71 they can provoke a StackOverflowException. They have direct access to the foreground executor because when you build a `GraphServiceClient` the object returned has a method `getExecutors()`. So that means that it's definitely part of the public API. At the moment `SynchronousExecutor` and `DefaultExecutors` are part of the public API too because we have not indicated to the user that they are  internal (we ourselves haven't decided yet!).

[...]

PR #52, opened 2018-02-15
improve SynchronousExecutor so recurses safely

davidmoten:
As discussed in #47, `SynchronousExecutor` is prone to stack overflow when the executor is called recursively. This PR fixes that by using `ThreadLocal` queues. 

Outstanding questions:

* do we want to null the `ThreadLocal` value every time the queue is emptied? This would make memory leak detectors happier (though this is not a leak of course) but would incur extra allocation costs when recursion was not happening.
* I assumed that the failure of one runnable in the queue should clear the queue (and this is tested). Make sense?

I've added a test that failed on the old code with a `StackOverflowException` and other tests for complete coverage of `SynchronousExecutor`.


davidmoten:
I'll redo this PR because rebase required and is from old repo

davidmoten:
Closing, will resubmit

Issue #40, opened 2018-01-30
GraphServiceClient rework


davidmoten:
The GraphServiceClient building api needs rework.

In particular, the builder allows you to do this:

```java
GraphServiceClient
  .builder()
  .fromConfig(config)
  .logger(logger)
  .buildClient();
```

However, the built client does not use the given logger in the `httpProvider` or `serializer` because they have already been instantiated before the `.logger(logger)` call.

To set the logger and the authenticationProvider I need to do this (and it depends on #38):

```java
        IClientConfig config = new DefaultClientConfig() {

            @Override
            public ILogger getLogger() {
                return logger;
            }

            @Override
            public IAuthenticationProvider getAuthenticationProvider() {
                return authProvider;
            }
        };

        IGraphServiceClient client = GraphServiceClient.builder() //
                .fromConfig(config) //
                .buildClient();
```

I'd prefer just to do this:

```java
GraphServiceClient client = GraphServiceClient
  .builder()
  .authenticationProvider(authProvider)
  .logger(logger)
  .buildClient();
```
Happy for this to be considered a bit later but should be sorted by public release.

dadheech115:
Hi @davidmoten, as your PR is already merged for this issue, can I go ahead and close this issue?

davidmoten:
Yes, thanks.

PR #60, opened 2018-02-28
rework GraphServiceClient

davidmoten:
See #40 

The crux of this PR is it supports this use of `GraphServiceClient`:

```java
client = GraphServiceClient.fromConfig(config);
```
and

```java
client = GraphServiceClient
  .builder()
  .logger(logger)
  .executors(executors)
  ...
  .buildClient();
```
It is not possible to override sensibly the behaviour of an instantiated config because it can only be overriden via inheritance.  For this reason the static factory method `fromConfig` returns a client straight away rather than allowing further builder modification.

The builder path is based on `DefaultClientConfig` and allows any of the bits below to be overriden in a sensible way. That is if I override the logger then it will be used in all the components.

*UML Dependency diagram - DefaultClientConfig*

deps

A quick note that there are more tweaks that could be done with `DefaultClientConfig` and the builder but I'll leave those for later PRs.

### DIFF
--- a/src/test/java/com/microsoft/graph/requests/extensions/GraphServiceClientTest.java
+++ b/src/test/java/com/microsoft/graph/requests/extensions/GraphServiceClientTest.java
@@ -1,0 +1,211 @@
+package com.microsoft.graph.requests.extensions;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Test;
+
+import com.microsoft.graph.authentication.IAuthenticationProvider;
+import com.microsoft.graph.concurrency.DefaultExecutors;
+import com.microsoft.graph.concurrency.ICallback;
+import com.microsoft.graph.concurrency.IExecutors;
+import com.microsoft.graph.concurrency.IProgressCallback;
+import com.microsoft.graph.core.ClientException;
+import com.microsoft.graph.core.DefaultClientConfig;
+import com.microsoft.graph.http.DefaultHttpProvider;
+import com.microsoft.graph.http.IHttpProvider;
+import com.microsoft.graph.http.IHttpRequest;
+import com.microsoft.graph.http.IStatefulResponseHandler;
+import com.microsoft.graph.logger.ILogger;
+import com.microsoft.graph.logger.LoggerLevel;
+import com.microsoft.graph.models.extensions.IGraphServiceClient;
+import com.microsoft.graph.serializer.DefaultSerializer;
+import com.microsoft.graph.serializer.ISerializer;
+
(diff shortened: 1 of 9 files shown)


cbales:
This separation is great. I agree it doesn't make sense to try to append methods to the `fromConfig` client. This is much cleaner and fixes many of the logger issues you highlighted.

src/test/java/com/microsoft/graph/requests/extensions/GraphServiceClientTest.java
```diff
@@ -0,0 +1,211 @@
...
+import com.microsoft.graph.logger.LoggerLevel;
+import com.microsoft.graph.models.extensions.IGraphServiceClient;
+import com.microsoft.graph.serializer.DefaultSerializer;
+import com.microsoft.graph.serializer.ISerializer;
+
+public class GraphServiceClientTest {
+
+    @Test
+    public void testClientMethodsReturnStuff() {
+        IGraphServiceClient client = GraphServiceClient.fromConfig(new DefaultClientConfig() {
```
cbales: These tests fail because AuthenticationProvider cannot be null. You can pass in an instance of `MockAuthenticationProvider` for testing purposes and `assertNotNull`

davidmoten: Ah some validation turned up that didn't exist before. The authentication provider must always be provided. I'm going to adjust the builder path to make this happen but also I suggest that we make the DefaultConfig.getAuthenticationProvider method abstract to force implementation. I'll update the PR and see what you think.

davidmoten:
I've updated the PR, ready for review again. 

The changes are:
* `DefaultClientConfig.getAuthenticationProvider` now abstract method to force implementation
*  Builder chaining used to force the specification of `authenticationProvider` in builder path
* Null checks performed on all builder methods
* More tests

So the builder now has to look like this:

```java
client = GraphServiceClient
    .builder()
    // must be specified first and enforced at compile time
    .authenticationProvider(authenticationProvider) 
    .logger(logger)
    ...
    .buildClient();
```
Given this situation I would normally remove the `builder()` method as below but I don't care much. Give me a yell if you'd like me to do that too. Result would be:

```java
client = GraphServiceClient
    .authenticationProvider(authenticationProvider) 
    .logger(logger)
    ...
    .buildClient();
```

davidmoten:
Ah, been merged to dev branch. I was just building a new version of the library and this PR wasn't present in master. Is that what you meant to do?

cbales:
Yes, I rebased your PR to dev now that the SDK is public. Dev is now the default branch. We will publish all PRs to dev. Once changes are ready to go live, we will push to master. PRs directly to master will be reserved for hotfixes, etc. Otherwise, master will match the latest release tag and published jar.

davidmoten:
Righto, thanks for the explanation.

PR #61, opened 2018-02-28
Enable more than 4GB for Travis CI

davidmoten:
These additional options in .travis.yml give 7.5GB of ram and should work now that we are on the public Travis infrastructure. They do work for me when I enable Travis on my fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,3 @@
 language: java
+dist: trusty
+sudo: required


cbales:
Unfortunately we will have to find a better long-term solution since this is getting close to the limits of the machines available on Travis. This SDK is a few months out of date with what is publicly in v1.0, with the most notable piece being Intune. Intune approximately quadruples the size of the library and requires a VM with around 16G of RAM to run CI. We will be investing internally in an Azure VM for this library, but the work is still in the backlog at the moment.

davidmoten:
May as well enable it now as it's building fine and checks PRs as well as master. Sure, do something different when the full API arrives (yikes). When are you thinking of bringing in the Intune stuff?

Would be helpful for me if you merge this PR because that way the PRs I put up through my fork go through my CI.

cbales:
This will be coming on board in a couple weeks at the latest.

deepak2016:
Hi David, wanted to check with you, how you enabled 7.5 GB ram for Travis CI. I took the commit in this pull request, but was not able to change heap size for jvm in travis.

davidmoten:
I got the info for the Travis changes from https://docs.travis-ci.com/user/reference/overview/.

You'll notice if you have 
```
sudo: required 
dist: trusty
```
then in theory the instance has 7.5GB.

deepak2016:
ok thanks for the information. I am trying these settings, however, travis CI is not going through.

PR #68, opened 2018-03-01
Functional test/sample code for chunked upload

cbales:
Several people were requesting samples on how to do this. The functional test also adds test coverage. Ported from [.NET client library functional test](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/tests/Microsoft.Graph.Test/Requests/Functional/OneDriveTests.cs).

I will also add to the wiki after this is pulled in.


src/test/java/com/microsoft/graph/functional/OneDriveTests.java
```diff
@@ -0,0 +1,80 @@
...
+     * @throws IOException if the input file is not found
+     * @throws InterruptedException if the chunked upload fails
+     */
+    @Test
+    public void testLargeFileUpload() throws IOException, InterruptedException {
+    	String itemId = "01BQHXQL5GQVAGCFJLYRH3EAG2YHGERMQA"; //Test upload folder
+    	
+    	//Get resource file from file system
+    	ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+    	InputStream uploadFile = classloader.getResourceAsStream("hamilton.jpg");
```
davidmoten: Do you need to get a specific classloader for this? Can't you just do

```java
InputStream uploadFile = OneDriveTests.class.getResourceAsStream("hamilton.jpg");
```

src/test/java/com/microsoft/graph/functional/OneDriveTests.java
```diff
@@ -0,0 +1,80 @@
...
+    			.post();
+    	
+    	ChunkedUploadProvider<DriveItem> chunkedUploadProvider = new ChunkedUploadProvider<DriveItem>(
+    			uploadSession, 
+    			testBase.graphClient, 
+    			uploadFile, 
+    			fileSize, 
+    			DriveItem.class);
+		
+    	chunkedUploadProvider.upload(null, callback);
```
davidmoten: Passing null for options is ugly (and the javadoc doesn't say it's nullable). Let's add an overload that omits the options parameter.

PR #72, opened 2018-03-06
Target Java 7

cbales:
The SDK must support Java versions >= 7 but was not set to compile to a specific version (so compiled at whatever version the developer had on their machine, in this case, Java 8).

- [4733365](https://github.com/microsoftgraph/msgraph-sdk-java/commit/4733365b19f8ba69efc6859884b7cc5ab454581f) - Compile to Java 7
- [52bdc57](https://github.com/microsoftgraph/msgraph-sdk-java/commit/52bdc5749e89213b8db1118af130dc5707100cac) - Resolve compiler error in JSE 7
- [5cc3ea7](https://github.com/microsoftgraph/msgraph-sdk-java/commit/5cc3ea7656b548525ca8477033ec9fe2d48a24f0) - remove reference to Java 8+ package


build.gradle
```diff
@@ -3,7 +3,7 @@
  *
  * This generated file contains a sample Java Library project to get you started.
  * For more details take a look at the Java Libraries chapter in the Gradle
- * user guide available at https://docs.gradle.org/3.5/userguide/java_library_plugin.html
+ * user guide available at https://docs.gradle.org/4.5/userguide/java_library_plugin.html
```
davidmoten: 4.5 -> 4.5.1

build.gradle
```diff
@@ -35,25 +35,16 @@ dependencies {
     compile 'com.sun.jersey:jersey-server:1.19.4'
 }
```
davidmoten: The publishing section is needed so I can call `./gradlew publishToMavenLocal` to generate maven artifacts locally. By all means use the properties ${mavenMajorVersion} etc. By the way in the dev branch gradle.properties should have `majorPatchVersion=0-SNAPSHOT`. It should only get the release version in `master` branch (if I have your workflow correct).

cbales: Added publishing section back in

build.gradle
```diff
@@ -35,6 +35,29 @@ dependencies {
...
+
+    publications {
+
+        maven(MavenPublication) {
+
+            groupId 'com.microsoft.graph'
+
+            artifactId 'microsoft-graph'
+
+            version '${mavenMajorVersion}.${mavenMinorVersion}.SNAPSHOT'
```
davidmoten: need the patch version too and should be `-SNAPSHOT`

build.gradle
davidmoten: Thanks. You might not be publishing snapshots or nightlies but any dev should be able to rebuild the release jar at the specificed version locally from whatever branch or tag without manual modification of build.gradle. For this reason the SNAPSHOT belongs in gradle.properties.

PR #73, opened 2018-03-06
Workload helper updates

cbales:
[12f6b8e](https://github.com/microsoftgraph/msgraph-sdk-java/commit/12f6b8e95c0963b61bcf28a5732f333e8e96a55d) - Add helper for multipart header and update OneNote multipart test to use header
[88f0cb6](https://github.com/microsoftgraph/msgraph-sdk-java/commit/88f0cb63bfe74c470af054296b1bcad6a16f3397) - Add default @odata.type parameters for Planner models, so the user does not need to specify them herself

### DIFF
--- a/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
+++ b/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
@@ -7,6 +7,8 @@ import java.io.InputStream;
 import java.math.BigInteger;
 import java.security.SecureRandom;
 
+import com.microsoft.graph.options.HeaderOption;
+
 /**
  * Helper for submitting multipart data
  * 
@@ -29,10 +31,18 @@ public class Multipart {
 	
 	/**
 	 * Get the multipart boundary for use in the request header
-	 * @return The multipart boundary
+	 * @return the multipart boundary
 	 */
 	public String boundary() {
 		return boundary;
+	}
+	
+	/**
+	 * Get the Content-Type header to send the multipart request
+	 * @return the multipart header option
+	 */
+	public HeaderOption header() {
+		return new HeaderOption("Content-Type", "multipart/form-data; boundary=\"" + boundary + "\"");
 	}
 	
 	/**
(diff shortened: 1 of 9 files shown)


src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -76,7 +86,7 @@ public void addHtmlPart(String name, String content) throws IOException {
 	}
 	
 	/**
-	 * Add an image part to the multipart body
+	 * Add an image/jpeg part to the multipart body
```
MIchaelMainer: Does OneNote only support adding jpegs? If so, let's rename this from addImagePart to addJpegPart. If it supports other types, we may want to consider adding an enum of supported image types.

cbales: I don't think that this helps much more than just using the addFilePart so I'll just remove it

MIchaelMainer: 👍

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -29,12 +31,20 @@ public Multipart() {
...
+	 * Get the Content-Type header to send the multipart request
+	 * @return the multipart header option
+	 */
+	public HeaderOption header() {
+		return new HeaderOption("Content-Type", "multipart/form-data; boundary=\"" + boundary + "\"");
+	}
+	
 	/**
 	 * Add a string part to the multipart body
 	 * @param name The name of the part
```
davidmoten: nit: fix case

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -75,16 +85,6 @@ public void addHtmlPart(String name, String content) throws IOException {
...
-	 * @param imageFile The image file
-	 * @throws IOException Throws an exception if the output stream cannot be written to
-	 */
-	public void addImagePart(String name, java.io.File imageFile) throws IOException {
-		addFilePart(name, "image/jpeg", imageFile);
-	}
-	
 	/**
 	 * Add a file part to the multipart body
 	 * @param name The name of the part
```
davidmoten: nit: fix case

PR #78, opened 2018-03-19
increase gradle heap requirement to 11GB

davidmoten:
as per title. 10GB works for linux.

Is the memory requirement larger on Windows?

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,7 +18,7 @@
 # org.gradle.parallel=true
 
 # The size of the library demands a large amount of RAM to build. Increase as necessary if you get GC errors
-org.gradle.jvmargs=-XX:MaxPermSize=512m -Xmx4g
+org.gradle.jvmargs=-XX:MaxPermSize=512m -Xmx10g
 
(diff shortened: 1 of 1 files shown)


davidmoten:
Increased heap size to 11G

davidmoten:
This is a trivial change that is required for running `./gradlew test` locally. Can this be merged please so I don't have to manually override it every time I want to test a PR.

deepak2016:
We are working to reduce the memory required to build Java SDK. Current memory requirement of ~11GB which is too large for the project. I am having discussions within the team regarding the same and will update this thread.

davidmoten:
>We are working to reduce the memory required to build Java SDK. Current memory requirement of ~11GB which is too large for the project. I am having discussions within the team regarding the same and will update this thread.

Glad to hear it. However, that will come well into the future. Let's merge this now because if someone clones this project at the moment and tries to build it with 4GB it runs forever due to the garbage collector going bananas.

deepak2016:
Agree to your point. Approving this commit, keeping reducing memory footprint as a goal.

Issue #44, opened 2018-02-08
Import wildcards


davidmoten:
I'd rank this as a nice to have. 

Import wildcards are used throughout the library. This has two primary effects:

* Namespace disambiguation must be controlled by class names rather than just package names
* Understanding the code outside a java enabled IDE is difficult (say in the browser)

The first one has been dealt with by ensuring no names are repeated between packages and I can cope with that though it is not ideal.

The second point is a more critical annoyance when trying to browse the code on github itself or as text locally because you can't see what package a referenced class is from.

The import wildcards in the generated code can be removed (and ambiguous names handled correctly with full qualification when necessary) using a technique like this (taken from https://github.com/davidmoten/state-machine/blob/master/state-machine-generator/src/main/java/com/github/davidmoten/fsm/model/Generator.java):

```java
Imports imports = new Imports();
ByteArrayOutputStream bytes = new ByteArrayOutputStream();
PrintStream out = new PrintStream(bytes);
out.format("package %s;\n", pkg);
out.println();
out.println("<IMPORTS>");
out.println();
out.format("public interface %s<Id> extends %s<%s, Id>{\n", behaviourClassSimpleName(),
                    imports.add(EntityBehaviour.class), imports.add(clsName));
...
result.print(new String(bytes.toByteArray(StandardCharsets.UTF_8))
            .replace("<IMPORTS>", imports.importsAsString()));
```
In a nutshell, you write your imports last after registering all types in the `Imports` and the `Imports` object decides which classes need to be fully qualified (first in best dressed).

An easy option is after pasting generated code, use an IDE like Eclipse to organize imports across the project (right mouse on project - Source - Organize imports).

Got an opinion on this one @cbales?

davidmoten:
@deepak2016 reopen this one please as import wildcards still in use.

baywet:
Hey @davidmoten 
Sorry for getting back to you this late on this one (I'm new on the team) and thanks for the suggestion!
I fixed this issue with version 2.0.0 as the usage of wildcard imports was slowing down compile time considerably for the beta version.
Closing :)

PR #96, opened 2018-06-13
Fixing import statments for request extension classes

deepak2016:
Fixes #
This results in memory gain of 2.5GB. Java SDK after this change builds with 7.5Gb RAM instead of taking more than 10GB

Fix import statements for all request extension classes in java sdk

### Other links
-


davidmoten:
See also #44

davidmoten:
From #44:
>An easy option is after pasting generated code, use an IDE like Eclipse to organize imports across the project (right mouse on project - Source - Organize imports).

This might save you a lot of trouble in 3.

deepak2016:
Yes, this is good idea, I totally agree with this approach. But we plan to automate SDK refresh process, hence planning to invest in fixing template files.

deepak2016:
I have tried that, java sdk builds with 1GB RAM, if I organize import for entire project.

davidmoten:
That's great! I don't mind organize imports being applied now on the project (I'll fix my PRs) because it will enable CI and allow general public to build the project (not everyone has 11GB on their machine!). I guess it depends on how long it takes to fix the templates.

deepak2016:
We are planning to be in a state to run Travis CI without issue by end of the month.

PR #97, opened 2018-06-14
fix published artifact versions

davidmoten:
This PR fixes  these things about publishing maven artifacts:
* the suffix (-SNAPSHOT) should be an optional part of the version and should be present in the dev branch always
* the publish section in build.gradle did not work because property substitution does not occur in single quotes. Needs double quotes.
* the publish section did not pick up the patch version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ publishing {
 
             artifactId 'microsoft-graph'
 
-            version '${mavenMajorVersion}.${mavenMinorVersion}-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


davidmoten:
This is a simple PR. Can we review and merge if ok? I'll move/copy the release process discussion stuff to an issue.

darrelmiller:
@deepak2016 Do you have any concerns about these changes?  To my untrained Java eye, they seem like simple goodness.

Issue #84, opened 2018-04-12
Allow to shutdown the default backgroundExecutor


J0s3f:
The default backgroundExecutor keeps running for about a minute after the last task until it is shut down by the JVM.

It would be good to have the option to shut it down programatically.

### Expected behavior

After all requests are handled the program quits if the main thread has ended.

### Actual behavior

Program keeps running for about a minute.

### Steps to reproduce the behavior

Just run the basic example given in the readme in a single main method.

### Possilble solution
Add a option to call shutdown() of the thread pool. The following works with reflection and gives the expected result. Obviously, a solution without reflection is needed.
```java
        //get background executor and shutdown:
        DefaultExecutors ex = (DefaultExecutors)graphClient.getExecutors();
        Field pool = ex.getClass().getDeclaredField("backgroundExecutor");
        pool.setAccessible(true);
        ThreadPoolExecutor executor = (ThreadPoolExecutor)pool.get(ex);
        System.out.println("Shutting down");
        executor.shutdown();
```
The most important part is to decide how the API/Interface should look. Implementing it should be trivial-

MIchaelMainer:
Thank you for reporting this.
753763

deepak2016:
This issue got fixed with PR# https://github.com/microsoftgraph/msgraph-sdk-java/pull/102

PR #102, opened 2018-07-11
Give the hook to shut down background executor gracefully

bullsseye:
Fixes # Fix for issue https://github.com/microsoftgraph/msgraph-sdk-java/issues/84

- Added a hook to shutdown background executor

### Other links
-

### DIFF
--- a/src/main/java/com/microsoft/graph/concurrency/DefaultExecutors.java
+++ b/src/main/java/com/microsoft/graph/concurrency/DefaultExecutors.java
@@ -145,7 +145,7 @@ public class DefaultExecutors implements IExecutors {
         return logger;
     }
     
-    public void shutDownBackgroundExecutors() {
+    public void shutdownBackgroundExecutors() {
     	this.backgroundExecutor.shutdown();
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/graph/concurrency/DefaultExecutors.java
```diff
@@ -144,5 +144,8 @@ public void run() {
     public ILogger getLogger() {
         return logger;
     }
-
+    
+    public void shutDownBackgroundExecutors() {
```
davidmoten: standard spelling is `shutdown` not `shutDown`

src/main/java/com/microsoft/graph/core/IBaseClient.java
```diff
@@ -86,4 +86,9 @@
      * Validates this client
      */
     void validate();
+    
+    /*
+     * Shuts down the executors.
+     */
+    void shutDown();
```
deepak2016: minor: spelling shutdown

davidmoten: yes I hoped my correction would be applied everywhere without having to say it!

Issue #95, opened 2018-06-11
DefaultHttpProvider body encoding problem


namjug-kim:
### Expected behavior

in DefaultHttpProvider#sendRequestInternal line 257

``` java
logger.logDebug("Sending " + serializable.getClass().getName() + " as request body");
final String serializeObject = serializer.serializeObject(serializable);
bytesToWrite = serializeObject.getBytes();
```

change body String to byte using `String#getBytes` it depends on `platform's default charset`

I use utf-8 and it may not work properly depending on the platform environment.

It would be better if could set up in DefaultClientConfig.

deepak2016:
Thanks @finalnova for reporting the issue. We will get back on this.

deepak2016:
@finalnova If I am understanding correctly, this issue would be present in all the instances where we are converting string to bytes?

deepak2016:
@finalnova APIs are working just fine irrespective of the encoding. I tried below encodings while doing getbytes for making api calls:

US-ASCII | Seven-bit ASCII, a.k.a. ISO646-US, a.k.a. the Basic Latin block of the Unicode character set
ISO-8859-1 | ISO Latin Alphabet No. 1, a.k.a. ISO-LATIN-1
UTF-8 | Eight-bit UCS Transformation Format
UTF-16 | Sixteen-bit UCS Transformation Format, byte order identified by an optional byte-order mark

finalnova:
@deepak2016 sorry. I guess I did not have enough explanation.
if filename is written in Korean (or other 3~4byte language) when trying to rename(PATCH request) DriveItem.name field is not always working fine.

``` java
DriveItem renameItem = new DriveItem();
renameItem.name = "테스트문서";

DriveItem patch = graphClient.me()
		.drive()
		.items(itemId)
		.buildRequest()
		.patch(renameItem);
```
If run the above code in two different encodings(ㅕUTF-8, ISO-8859-1), the result is:

```
-Dfile.encoding=UTF-8
  - 200 OK

-Dfile.encoding=ISO-8859-1
  - 400 : Bad Request 
```

deepak2016:
@namjug-kim I am able to reproduce the issue now. Thanks for putting up an example.
Regarding the fix, I'm thinking if encoding should always be UTF-8 or is there any scenario where we would like it to be configurable?

namjug-kim:
@deepak2016 In my case, UTF-8 worked fine in all situations. I think it's okay to fix UTF-8.
[...]

PR #109, opened 2018-07-31
Utf issue fix

NakulSabharwal:
Fixing issue https://github.com/microsoftgraph/msgraph-sdk-java/issues/95 by giving UTF-8 as parameter to getBytes.

### DIFF
--- a/src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
@@ -414,7 +414,7 @@ public class DefaultHttpProvider implements IHttpProvider {
         	in = new ByteArrayInputStream("{}".getBytes(ENCODING_TYPE));
         }
         catch(UnsupportedEncodingException ex) {
-        	ex.printStackTrace();
+        	logger.logError(ex.getMessage(), ex);
         }
         
     	return handleJsonResponse(in, responseHeaders, clazz);
--- a/src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
@@ -407,7 +407,8 @@ public class DefaultHttpProvider implements IHttpProvider {
      * @param clazz           the type of the response object
      * @return                the JSON object
      */
-    private <Result> Result handleEmptyResponse(Map<String, List<String>> responseHeaders, final Class<Result> clazz) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
```diff
@@ -401,9 +407,17 @@ private InputStream handleBinaryStream(final InputStream in) {
...
+    private <Result> Result handleEmptyResponse(Map<String, List<String>> responseHeaders, final Class<Result> clazz) 
+    		throws UnsupportedEncodingException{
     	//Create an empty object to attach the response headers to
-        InputStream in = new ByteArrayInputStream("{}".getBytes());
+    	InputStream in = null;
+    	try{
+        	in = new ByteArrayInputStream("{}".getBytes(ENCODING_TYPE));
+        }
+        catch(UnsupportedEncodingException ex) {
+        	logger.logError(ex.getMessage(), ex);
```
deepak2016: Instead of logging error here, maybe you should create new catch block at the end of sendRequestInternal method. There is already few catch blocks which will log incorrectly in this particular case.

src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
```diff
@@ -58,6 +59,11 @@
      * The content type for JSON responses
      */
     static final String JSON_CONTENT_TYPE = "application/json";
+    
+    /**
+     * The encoding type for getBytes
+     */
+    static final String ENCODING_TYPE = "UTF-8";
```
davidmoten: rename to JSON_ENCODING or JSON_CHARSET

src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java
```diff
@@ -341,6 +347,11 @@ public void run() {
             final boolean shouldLogVerbosely = logger.getLoggingLevel() == LoggerLevel.DEBUG;
             logger.logError("Graph service exception " + ex.getMessage(shouldLogVerbosely), ex);
             throw ex;
+        } catch (final UnsupportedEncodingException ex) {
+        	final ClientException clientException = new ClientException("Unsupported encoding",
+                    ex);
+            logger.logError("Unsupported encoding exception " + ex.getMessage(), ex);
```
davidmoten: nit: Should probably be "Unsupported encoding problem: "  (with colon) and could apply to other `logError` calls in this class.

PR #110, opened 2018-08-02
Fixes #105: Explicitly set getBytes encoding

### LINKED ISSUE #105: getBytes encoding should be explicit
Extending discussion in #95 there are many places in the project source where platform dependent character encoding is used via a call to `getBytes()` instead of for instance `getBytes(utf8)` (or whatever is appropriate). All of these calls should use an explicit `Charset`.

A quick grep showed these:

```
./src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java:                    bytesToWrite = serializeObject.getBytes();
./src/main/java/com/microsoft/graph/http/DefaultHttpProvider.java:        InputStream in = new ByteArrayInputStream("{}".getBytes());
./src/main/java/com/microsoft/graph/models/extensions/Multipart.java:		addPart(name, contentType, content.getBytes());
./src/main/java/com/microsoft/graph/models/extensions/Multipart.java:		out.write(partContent.getBytes());
./src/main/java/com/microsoft/graph/models/extensions/Multipart.java:		out.write(returnContent.getBytes());
./src/main/java/com/microsoft/graph/models/extensions/Multipart.java:		finalStream.write(addEnding().getBytes());
./src/test/java/com/microsoft/graph/http/MockConnection.java:        return new ByteArrayInputStream(mData.getJsonResponse().getBytes());
./src/test/java/com/microsoft/graph/functional/SharePointTests.java://        String url = "u!" + Base64.getUrlEncoder().encodeToString(testSite.webUrl.getBytes());
./src/test/java/com/microsoft/graph/functional/OneNoteTests.java:        byte[] pageStream = content.getBytes();
```

NakulSabharwal:
Fixes #105: Explicitly set getBytes encoding

### DIFF
--- a/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
+++ b/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
@@ -19,6 +19,10 @@ import com.microsoft.graph.options.HeaderOption;
 public class Multipart {
 	private String boundary;
 	private static final String RETURN = "\r\n";
+	/**
+     * The encoding type for getBytes
+     */
+    static final String JSON_ENCODING = "UTF-8";
 	private ByteArrayOutputStream out;
 	
 	/**
@@ -53,7 +57,7 @@ public class Multipart {
 	 * @throws IOException Throws an exception if the output stream cannot be written to
 	 */
 	public void addPart(String name, String contentType, String content) throws IOException {
-		addPart(name, contentType, content.getBytes());
+		addPart(name, contentType, content.getBytes(JSON_ENCODING));
 	}
 	
 	/**
@@ -69,10 +73,10 @@ public class Multipart {
 				"Content-Disposition:form-data; name=\"" + name + "\"" + RETURN +
 				"Content-Type:" + contentType + RETURN +
 				RETURN;
-		out.write(partContent.getBytes());
+		out.write(partContent.getBytes(JSON_ENCODING));
 		out.write(byteArray);
 		String returnContent = RETURN + RETURN;
-		out.write(returnContent.getBytes());
+		out.write(returnContent.getBytes(JSON_ENCODING));
 	}
 	
 	/**
@@ -121,7 +125,7 @@ public class Multipart {
 	 */
 	public byte[] content() throws IOException {
 		ByteArrayOutputStream finalStream = out;
-		finalStream.write(addEnding().getBytes());
+		finalStream.write(addEnding().getBytes(JSON_ENCODING));
 		return finalStream.toByteArray();
 	}
 	
--- a/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
+++ b/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
@@ -20,9 +20,17 @@ public class Multipart {
 	private String boundary;
 	private static final String RETURN = "\r\n";
 	/**
-     * The encoding type for getBytes
+     * The text encoding type for getBytes
      */
-    static final String JSON_ENCODING = "UTF-8";
+    static final String TEXT_ENCODING = "UTF-8";
+    /**
+     * The file encoding type in getBytes
+     * Encoding used for files{image, pdf, etc} is UTF-8 because
+     * https://developer.microsoft.com/en-us/graph/docs/concepts/onenote_images_files#add-a-file-attachment
+     * mentions "send the binary data, don't use Base64 or otherwise encode it"
+     * Also, https://en.wikipedia.org/wiki/MIME#Multipart_messages explains 7bit, 8bit or binary like UTF-8.
+     */
+    static final String FILE_ENCODING = "UTF-8";
 	private ByteArrayOutputStream out;
 	
 	/**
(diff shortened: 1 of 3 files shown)


src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -291,7 +296,7 @@ public void testGetPreview() {
      * @throws InterruptedException
```
deepak2016: edit comment now that it is throwing extra exception

davidmoten:
Please change issue title (e.g. 'Explicity set getBytes encoding'), current title is a bit crazy

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -19,6 +19,10 @@
 public class Multipart {
 	private String boundary;
 	private static final String RETURN = "\r\n";
+	/**
+     * The encoding type for getBytes
+     */
+    static final String JSON_ENCODING = "UTF-8";
```
davidmoten: The content is not necessarily JSON! You need to investigate what the encoding is for the Multipart content. I'm assuming this is for SMTP Multipart data.  Please justify your choice of encoding (and put it in a comment in the source).

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -49,6 +50,10 @@
     private OnenotePage testPage;
     private OnenoteSection testSection;
     private SectionGroup testSectionGroup2;
+    /**
+     * The encoding type for getBytes
+     */
+    static final String JSON_ENCODING = "UTF-8";
```
davidmoten: Again, investigate what encoding is appropriate. The field name JSON_ENCODING suggests you haven't though about it.

src/test/java/com/microsoft/graph/http/MockConnection.java
```diff
@@ -18,6 +18,10 @@
     private final ITestConnectionData mData;
     private HashMap<String, String> mHeaders = new HashMap<>();
     private Boolean mFollowRedirects;
+    /**
+     * The encoding type for getBytes
+     */
+    static final String JSON_ENCODING = "UTF-8";
```
davidmoten: This time we are fine, because you are encoding a JsonResponse

davidmoten:
Please look at conventions in wording of PRs in this project. I would call it "Explicitly set getBytes encoding" and then in the start of the first comment say "Fixes #105".

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: I have updated String names. Also, https://developer.microsoft.com/en-us/graph/docs/concepts/onenote_images_files#add-a-file-attachment mentions "send the binary data, don't use Base64 or otherwise encode it" and https://en.wikipedia.org/wiki/MIME#Multipart_messages explains to use encoding 7bit, 8bit or binary like UTF-8.

src/test/java/com/microsoft/graph/http/MockConnection.java
NakulSabharwal: Ok, Keeping this as JSON_ENCODING.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
NakulSabharwal: Changed it to HTML_ENCODING. The W3C recommends UTF-8 as the default encoding in XML and HTML (reference : https://en.wikipedia.org/wiki/UTF-8).

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
davidmoten: The job does not seem to be done yet. Can the user set the Content-Transfer-Encoding MIME header (which you would then have to honour)?

davidmoten:
@NakulSabharwal  I meant put Fixes #105 in the first comment (which is editable by you) not in the title.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
darrelmiller: @davidmoten Currently we don't appear to support setting any headers for parts other than Content-Type and Content-Transfer-Encoding.  If we did then the user would be highly recommended to pass the properly encoded bytes in as a byte array.

@NakulSabharwal For the overload that accepts a name and creates a Content-Disposition header, we can't really just use UTF-8 decoding on that' string.  MIME headers in multipart only support US-ASCII and percent encoding.  The full spec for multipart/form-data is here https://tools.ietf.org/html/rfc7578 and it was updated fairly recently.

We should get together and consider what multipart features we need to support for existing Graph APIs and write up requirements across all the languages.  Assuming UTF-8 strings for JSON is fine, but it's not for HTML.  We also need to look into whether we need to support filename as a content disposition parameter.

Handling multipart properly is tricky, and I want to make sure we are doing this consistently across all the languages and that we support all the features we need for Graph.

Unfortunately I'm out this week, but maybe we can gather some requirements and meet next week to nail down exactly what features we are trying to support, that would be great.  /cc @deepak2016

PR #114, opened 2018-08-16
Unit test to verify multi-level additional data setter

### LINKED ISSUE #113: PR #104 work incomplete
This is issue is to ensure that changes requested on an already merged PR #104 are not forgotten.
In particular, no unit test was supplied and an error was logged not thrown which should be justified in comments at least (or amended to rethrow the error).
Can we get the comments made on that PR addressed please @dadheech115.
[AB#6036](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6036)

dadheech115:
Added a unit test case to verify multi-level additional data setter and added a comment to explain the IllegalArgumentException

Fixes #113 

-Have Added a test function in AddtionalDataTests.java class.
-Have also added a comment in DefaultSerializer which explains the reason for not throwing IllegalArgumentException


src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -148,7 +148,8 @@ else if (fieldObject != null && fieldObject instanceof IJsonBackedObject) {
                     setChildAdditionalData((IJsonBackedObject) fieldObject,rawJson.get(field.getName()).getAsJsonObject());
                 }
             } catch (IllegalArgumentException | IllegalAccessException e) {
-                logger.logError("Unable to access child fields of " + serializedObject.getClass().getSimpleName(), e);
+                //Not throwing the IllegalArgumentException as the Serialized Object would still be usable even if the additional data is not set.
```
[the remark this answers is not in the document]
darrelmiller: I would argue that just because the Java reflection engine hands us a field that it won't let us access, isn't a good enough reason to throw an exception at the user.  Especially when the only reason we are reflecting over it is because we are hunting for extra properties that might be there so that we can fill the additionalProperties collection.  

I do agree that the try/catch is a bit too wide of a net and we should try and limit the lines we are actually swallowing exceptions for.

However, this brings me to my main concern here.  If I understand this code, we are reflecting over every single property of every object that is returned from the Graph.  This seems like an expensive solution to a problem that by definition is an exception case.  The .Net library has the luxury of doing this because the deserializer library builds the list of unknown properties while it is parsing the JSON.  

I really don't like the idea of running this chunk of reflection code on the retrieval of every single object.  That just sounds like asking for trouble.

davidmoten:
I notice also that the setChildAdditionalData is recursive. This sets us up for an infinite recursion if object A has a reference to B and B has a reference to A as a simple example. Do we need protection for this and if not why not?

davidmoten:
I'm also concerned by
* is it even appropriate to recursively set the additional data because we are not traversing the JSON as we traverse the object structure
* fields that have already been set will be overwritten?

Again, forgive the superficial review!

dadheech115:
@davidmoten , as for your concern about infinite recursion. As we are recursively traversing the JSON object returned from Graph server, there is no chance of one JSON object being referred from a sub-element. Hence there will not be any infinite recursion scenario.
And if I understand it correctly, you want to know why are we not traversing the json at the time we are deserializing other properties of a Model. So, the reason for that is we are using GSON library for object parsing. So instead of modifying the code of that library to include our logic of setting additional data, we thought it would be wise to once deserialize the object with all the properties available using this library [so not to reinvent the wheel] and then traverse and see if there are any additional values that have been left out from this parser.

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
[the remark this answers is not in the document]
dadheech115: @davidmotel,I agree that 100% code coverage is the ideal test case result that we should strive for. And we will further try to enhance this test case to cover most of the branching in the code.
As for the exceptions caught at the end of setting up additional data, these exceptions are thrown by field.get(object) method of  java.lang.reflect.Field class. So, we are just catching it and not throwing it to break the flow.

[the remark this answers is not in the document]
davidmoten: No, you are catching IllegalArgumentException and IllegalAccessException. A NullPointerException will be thrown not caught. This is why unit tests are important.

davidmoten:
>@davidmoten , as for your concern about infinite recursion. As we are recursively traversing the JSON object returned from Graph server, there is no chance of one JSON object being referred from a sub-element. Hence there will not be any infinite recursion scenario.

Ok, thanks.

>And if I understand it correctly, you want to know why are we not traversing the json at the time we are deserializing other properties of a Model. So, the reason for that is we are using GSON library for object parsing. So instead of modifying the code of that library to include our logic of setting additional data, we thought it would be wise to once deserialize the object with all the properties available using this library [so not to reinvent the wheel] and then traverse and see if there are any additional values that have been left out from this parser.

Ok, so I want to check this scenario. If 

* the object graph A -> B is deserialized (we end up with an object A that points to B)
* and the json fragment for A has an unmapped property x = y (additional data)
* and the json fragment for B has an unmapped property x = z (additional data)

Will the additional data for deserialized B end up with x = y instead of x = z?

Another scenario is 

* the object graph A -> B is deserialized (we end up with an object A that points to B)
* and the json fragment for A has an unmapped property x = y (additional data)
* and the json fragment for B has no unmapped properties

Will the additional data for deserialized B end up with x = y?

If these are valid scenarios I'd like the desired outcomes to be unit tested.

davidmoten:
Until the scenarios I mention above have been addressed I think the original PR should be backed out of dev branch (and master if there).

deepak2016:
@davidmoten thanks for bringing this up, we will get working on fixing this issue.

baywet:
Hey @davidmoten 
I just did some light refactoring to improve readability and robustness.
I don't want to touch too much this section of the code as I'm still getting familiar with the code base at this point.
I believe most if not all the comments you had have been addressed in previous PRs.
Let me know if I missed anything.

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -118,59 +118,70 @@ public DefaultSerializer(final ILogger logger) {
...
-                            if (element instanceof IJsonBackedObject) {
-                                final JsonElement elementRawJson = rawJsonArray.get(i);
-                                setChildAdditionalData((IJsonBackedObject) element, elementRawJson.getAsJsonObject());
+                        // If the object is a list of Graph objects, iterate through elements
+                        else if (fieldObject instanceof List) {
+                            final JsonElement collectionJson = rawJson.get(field.getName());
+                            final List<?> fieldObjectList = (List<?>) fieldObject;
+                            if (collectionJson != null && collectionJson.isJsonArray()) {
+                                final JsonArray rawJsonArray = (JsonArray) collectionJson;
+                                if(rawJsonArray.size() == fieldObjectList.size()) {
```
MIchaelMainer, line 157: I can't fathom a scenario when this isn't true. But if it were ever false, we should log this as potential missing additional data. 

`rawJsonArray.size() == fieldObjectList.size()`

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -118,59 +118,70 @@ public DefaultSerializer(final ILogger logger) {
...
-                // If the object is a HashMap, iterate through its children
-                if (fieldObject instanceof HashMap) {
-                    HashMap<String, Object> serializableChildren = (HashMap<String, Object>) fieldObject;
-                    Iterator<Entry<String, Object>> it = serializableChildren.entrySet().iterator();
+        if(rawJson != null) {
+            for (java.lang.reflect.Field field : serializedObject.getClass().getFields()) {
+                try {
+                    if(field != null) {
+                        final Object fieldObject = field.get(serializedObject);
+                        if (fieldObject instanceof HashMap) {
```
MIchaelMainer, line 128: Add a test for `fieldObject instanceof HashMap` using `serializer.deserializeObject(sourcestring, class);`

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -118,59 +118,70 @@ public DefaultSerializer(final ILogger logger) {
...
-                    final List<?> fieldObjectList = (List<?>) fieldObject;
-                    if (collectionJson.isJsonArray() && ((JsonArray)collectionJson).size() == fieldObjectList.size()) {
-                        final JsonArray rawJsonArray = (JsonArray) collectionJson;
-                        for (int i = 0; i < fieldObjectList.size(); i++) {
-                            final Object element = fieldObjectList.get(i);
-                            if (element instanceof IJsonBackedObject) {
-                                final JsonElement elementRawJson = rawJsonArray.get(i);
-                                setChildAdditionalData((IJsonBackedObject) element, elementRawJson.getAsJsonObject());
+                        // If the object is a list of Graph objects, iterate through elements
+                        else if (fieldObject instanceof List) {
```
MIchaelMainer, line 152: Add a test for `fieldObject instanceof HashMap` using `serializer.deserializeObject(sourcestring, class);`

One test for a homogenous list of objects, and one for a hetergenous list of objects.

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -118,59 +118,70 @@ public DefaultSerializer(final ILogger logger) {
...
 
-                        // If the item is a valid Graph object, set its additional data
-                        if (child instanceof IJsonBackedObject) {
-                            AdditionalDataManager childAdditionalDataManager = ((IJsonBackedObject) child).additionalDataManager();
-                            if(rawJson != null && field != null && rawJson.get(field.getName()) != null && rawJson.get(field.getName()).isJsonObject() 
-                            		&& rawJson.get(field.getName()).getAsJsonObject().get(pair.getKey()).isJsonObject()) {
-                            	childAdditionalDataManager.setAdditionalData(rawJson.get(field.getName()).getAsJsonObject().get(pair.getKey()).getAsJsonObject());
-                            	setChildAdditionalData((IJsonBackedObject) child,rawJson.get(field.getName()).getAsJsonObject().get(pair.getKey()).getAsJsonObject());
+                                // If the item is a valid Graph object, set its additional data
+                                if (child instanceof IJsonBackedObject) {
```
MIchaelMainer, line 139: I'd like to see a test in case where the child is a OData primitive, do we add non-schematized fields to additionalData?

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
baywet: fixed

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
```diff
@@ -118,59 +118,70 @@ public DefaultSerializer(final ILogger logger) {
...
+                            final JsonElement fieldElement = rawJson.get(field.getName());
+                            if(fieldElement != null && fieldElement.isJsonObject()) {
+                                childAdditionalDataManager.setAdditionalData(fieldElement.getAsJsonObject());
+                                setChildAdditionalData((IJsonBackedObject) fieldObject,fieldElement.getAsJsonObject());
                             }
                         }
                     }
+                } catch (IllegalArgumentException | IllegalAccessException e) {
+                    //Not throwing the IllegalArgumentException as the Serialized Object would still be usable even if the additional data is not set.
+                    logger.logError("Unable to set child fields of " + serializedObject.getClass().getSimpleName(), e);
```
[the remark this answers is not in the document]
baywet: fixed

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
baywet: added a unit test. We currently don't have a case where hashmaps have heterogeneous values and I didn't want to add too much fake code in the test setup.

src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
baywet: tweaked the last unit test I added so it tests for that as well.

MIchaelMainer (approved):

PR #131, opened 2018-09-10
Getbytes encoding fix

### LINKED ISSUE #105: getBytes encoding should be explicit

NakulSabharwal:
Fixes #105 

- Set getbytes in Multipart to US-ASCII.
- Set getbytes in jsonResponse to UTF-8.
- Added getter and setter for boundary and  Content-Type of Multipart.
- Removed function addPart(String name, String contentType, String content).
- Renamed function addPart(String name, String contentType, byte[] byteArray) To addFormData(String name, String contentType, byte[] byteArray).
- Added function addPart(Map<String, String> headers, byte[] content) for user provided headers and byte content.

### Other links
-https://tools.ietf.org/html/rfc7578

### DIFF
--- a/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
+++ b/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
@@ -74,6 +74,75 @@ public class Multipart {
 		return new HeaderOption("Content-Type", contentType + "; boundary=\"" + boundary + "\"");
 	}
 	
+	private void writePartData(String partContent, byte[] byteArray) throws IOException{
+		out.write(partContent.getBytes(MULTIPART_ENCODING));
+		out.write(byteArray);
+		String returnContent = RETURN + RETURN;
+		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+	}
+	
+	private String createPartHeader(Map<String, String> headers, String name, String contentType, String filename) {
+		String partContent = addBoundary();
+		if(headers != null) {
+			List<String> listContentDisposition = Arrays.asList("filename","creation-date","modification-date","read-date",
+					"size","name","voice","handling","preview-type");			
+			List<String> mainHeaders = Arrays.asList("Content-Disposition","Content-Type","charset");
+			
+			if(headers.containsKey("Content-Disposition")) {			
+				partContent += "Content-Disposition:"+headers.get("Content-Disposition")+";"; 
+				for (Map.Entry<String,String> entry : headers.entrySet()) {
+					if(listContentDisposition.contains(entry.getKey())) {
+						partContent += " " + entry.getKey() + "=\"" + entry.getValue() + "\";";
+					}
+				}
+				partContent = partContent.substring(0, partContent.length()-1);
+				partContent += RETURN;
+				
+			}
+			
+			if(headers.containsKey("Content-Type")) {
+				partContent += "Content-Type:"+headers.get("Content-Type")+";"; 
+				if(headers.containsKey("charset")) {
+					partContent += "charset=\"" + headers.get("charset") + "\";"; 
+				}
+				partContent = partContent.substring(0, partContent.length()-1);
+				partContent += RETURN;
+			}
+			
+			for(Map.Entry<String,String> entry : headers.entrySet()) {
+				if(mainHeaders.contains(entry.getKey())==false && listContentDisposition.contains(entry.getKey())==false) {
+					partContent += entry.getKey() +":"+entry.getValue() + RETURN;
+				}
+			}
+			partContent += RETURN;
+		}
+		else if(filename != null && name != null) {
+			partContent +=
+					"Content-Disposition:form-data; name=\"" + name + "\"" + "; filename=\"" + filename + "\"" + RETURN +
+					"Content-Type:" + contentType + RETURN +
+					RETURN;
+		}
+		else if(filename != null) {
+			partContent +=
+					"Content-Disposition:form-data; filename=\"" + filename + "\"" + RETURN +
+					"Content-Type:" + contentType + RETURN +
+					RETURN;
+		}
+		else if(name != null){
+			partContent +=
+					"Content-Disposition:form-data; name=\"" + name + "\"" + RETURN +
+					"Content-Type:" + contentType + RETURN +
+					RETURN;
+		}
+		else {
+			partContent +=
+					"Content-Disposition:form-data;" + RETURN +
+					"Content-Type:" + contentType + RETURN +
+					RETURN;
+		}		
+		return partContent;
+	}
+	
 	/**
 	 * Add a part to the multipart body
 	 * @param name The name of the part
@@ -82,16 +151,8 @@ public class Multipart {
 	 * @throws IOException Throws an exception if the output stream cannot be written to
 	 */
 	public void addFormData(String name, String contentType, byte[] byteArray) throws IOException {
-		String partContent = addBoundary();
-		partContent +=
-				"Content-Disposition:form-data; name=\"" + name + "\"" + RETURN +
-				"Content-Type:" + contentType + RETURN +
-				RETURN;
-		out.write(partContent.getBytes(MULTIPART_ENCODING));
-		out.write(byteArray);
-		String returnContent = RETURN + RETURN;
-		out.write(returnContent.getBytes(MULTIPART_ENCODING));
-		System.out.println(partContent);
+		String partContent = createPartHeader(null, name, contentType, null);
+		writePartData(partContent, byteArray);
 	}
 	
 	/**
@@ -101,15 +162,8 @@ public class Multipart {
 	 * @throws IOException Throws an exception if the output stream cannot be written to
 	 */
 	public void addPart(String contentType, byte[] byteArray) throws IOException {
-		String partContent = addBoundary();
-		partContent +=
-				"Content-Disposition:form-data;" + RETURN +
-				"Content-Type:" + contentType + RETURN +
-				RETURN;
-		out.write(partContent.getBytes(MULTIPART_ENCODING));
-		out.write(byteArray);
-		String returnContent = RETURN + RETURN;
-		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+		String partContent = createPartHeader(null, null, contentType, null);
+		writePartData(partContent, byteArray);
 	}
 	
 	/**
@@ -132,16 +186,8 @@ public class Multipart {
 	public void addFilePart(String name, String contentType, java.io.File file) throws IOException {
 		InputStream fileStream = new FileInputStream(file);
 		byte[] fileBytes = getByteArray(fileStream);
-		String partContent = addBoundary();
-		partContent +=
-				"Content-Disposition:form-data; name=\"" + name + "\"" + "; filename=\"" + file.getName() + "\"" + RETURN +
-				"Content-Type:" + contentType + RETURN +
-				RETURN;
-		System.out.println(partContent);
-		out.write(partContent.getBytes(MULTIPART_ENCODING));
-		out.write(fileBytes);
-		String returnContent = RETURN + RETURN;
-		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+		String partContent = createPartHeader(null, name, contentType, file.getName());
+		writePartData(partContent, fileBytes);
 	}
 	
 	/**
@@ -151,44 +197,8 @@ public class Multipart {
 	 * @throws IOException Throws an exception if the output stream cannot be written to
 	 */
 	public void addPart(Map<String, String> headers, byte[] content) throws IOException{
-		String partContent = addBoundary();
-		List<String> listContentDisposition = Arrays.asList("filename","creation-date","modification-date","read-date",
-				"size","name","voice","handling","preview-type");			
-		List<String> mainHeaders = Arrays.asList("Content-Disposition","Content-Type","charset");
-		
-		if(headers.containsKey("Content-Disposition")) {			
-			partContent += "Content-Disposition:"+headers.get("Content-Disposition")+";"; 
-			for (Map.Entry<String,String> entry : headers.entrySet()) {
-				if(listContentDisposition.contains(entry.getKey())) {
-					partContent += " " + entry.getKey() + "=\"" + entry.getValue() + "\";";
-				}
-			}
-			partContent = partContent.substring(0, partContent.length()-1);
-			partContent += RETURN;
-			
-		}
-		
-		if(headers.containsKey("Content-Type")) {
-			partContent += "Content-Type:"+headers.get("Content-Type")+";"; 
-			if(headers.containsKey("charset")) {
-				partContent += "charset=\"" + headers.get("charset") + "\";"; 
-			}
-			partContent = partContent.substring(0, partContent.length()-1);
-			partContent += RETURN;
-		}
-		
-		for(Map.Entry<String,String> entry : headers.entrySet()) {
-			if(mainHeaders.contains(entry.getKey())==false && listContentDisposition.contains(entry.getKey())==false) {
-				partContent += entry.getKey() +":"+entry.getValue() + RETURN;
-			}
-		}
-		
-		System.out.println(partContent);
-		partContent += RETURN;
-		out.write(partContent.getBytes(MULTIPART_ENCODING));
-		out.write(content);
-		String returnContent = RETURN + RETURN;
-		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+		String partContent = createPartHeader(headers, null, null, null);
+		writePartData(partContent, content);
 	}
 	
 	/**
--- a/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
+++ b/src/main/java/com/microsoft/graph/models/extensions/Multipart.java
@@ -81,19 +81,22 @@ public class Multipart {
 	
 	private String createPartHeader(String name, String contentType, String filename) {
         String partContent = addBoundary();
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,46 +38,132 @@ public Multipart() {
...
+	
+	private void writePartData(String partContent, byte[] byteArray) throws IOException{
+		out.write(partContent.getBytes(MULTIPART_ENCODING));
+		out.write(byteArray);
+		String returnContent = RETURN + RETURN;
+		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+	}
+	
+	private String createPartHeader(Map<String, String> headers, String name, String contentType, String filename) {
+		String partContent = addBoundary();
```
davidmoten: Nit: make `partContent` a `StringBuilder` to avoid unnecessary copying

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,46 +38,132 @@ public Multipart() {
...
+	}
+	
+	private void writePartData(String partContent, byte[] byteArray) throws IOException{
+		out.write(partContent.getBytes(MULTIPART_ENCODING));
+		out.write(byteArray);
+		String returnContent = RETURN + RETURN;
+		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+	}
+	
+	private String createPartHeader(Map<String, String> headers, String name, String contentType, String filename) {
```
davidmoten: Make this method static if doesn't use parent object state

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,46 +38,132 @@ public Multipart() {
...
+			List<String> mainHeaders = Arrays.asList("Content-Disposition","Content-Type","charset");
+			
+			if(headers.containsKey("Content-Disposition")) {			
+				partContent += "Content-Disposition:"+headers.get("Content-Disposition")+";"; 
+				for (Map.Entry<String,String> entry : headers.entrySet()) {
+					if(listContentDisposition.contains(entry.getKey())) {
+						partContent += " " + entry.getKey() + "=\"" + entry.getValue() + "\";";
+					}
+				}
+				partContent = partContent.substring(0, partContent.length()-1);
```
davidmoten: You want to strip the final semicolon. Why's that?

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
davidmoten: Please add unit test coverage of this method.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
darrelmiller: Content-Disposition parameters are prefixed by semi-colon, so removing the trailing one is technically correct.

However, I'm a little puzzled as to this approach where we pretend Content-Disposition parameters are headers and then we lift them out of the headers collection to construct the Content-Disposition header value.  The list of headers should just be a list of HTTP headers where the content-disposition value is pre-constructed.  If we need a helper to build that value, then let's build a createContentDispositionValue() function.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -82,7 +173,7 @@ public void addPart(String name, String contentType, byte[] byteArray) throws IO
 	 * @throws IOException Throws an exception if the output stream cannot be written to
 	 */
 	public void addHtmlPart(String name, String content) throws IOException {
-		addPart(name, "text/html", content);
+		addFormData(name, "text/html", content.getBytes(MULTIPART_ENCODING));
```
darrelmiller: We can't make assumptions about the encoding of HTML content.  If we keep this addHtmlPart method, it needs to accept a byte array and not a string.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,46 +38,132 @@ public Multipart() {
...
+			for(Map.Entry<String,String> entry : headers.entrySet()) {
+				if(mainHeaders.contains(entry.getKey())==false && listContentDisposition.contains(entry.getKey())==false) {
+					partContent += entry.getKey() +":"+entry.getValue() + RETURN;
+				}
+			}
+			partContent += RETURN;
+		}
+		else if(filename != null && name != null) {
+			partContent +=
+					"Content-Disposition:form-data; name=\"" + name + "\"" + "; filename=\"" + filename + "\"" + RETURN +
```
darrelmiller: There needs to be a space after the colon in the Content-Disposition header.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -429,6 +434,59 @@ public void testMultipartPost(){
             fail("Unable to write to output stream");
         }
     }
+    
+    /**
+     * Test posting multipart content to a page
+     */
+    @Test
+    public void testMultipartHeadersMapPost(){
```
deepak2016: rename to testMultipartPostWithHeadersMap

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Using String.format(...) now for building String

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: This method cannot be made static because it uses boundary string which is dependent on parent object state.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: - I have now changed the way in developer will use this addPart(Map<String, String> headers, byte[] content) and added Helper function String createContentHeaderValue(String contentValue, Map<String, String> contentDispParameter). 
- contentDispParameter map will contain (key=parameterName, Value=parameterValue). eg. (key="name", value="name-val").
- createContentHeaderValue helper function will construct the string "contentValue;name="nameval";..." which will be then passed in headers map as (key=Content-Disposition, value=string from helper function), (key=Content-Type, value=string from helper).

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Changed addHtmlPart to accept bytes.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Added space after the colon in the Content-Disposition header.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
NakulSabharwal: Renamed unit test function to testMultipartPostWithHeadersMap.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,56 +36,158 @@ public Multipart() {
...
+	
+	private void writePartData(String partContent, byte[] byteArray) throws IOException{
+		out.write(partContent.getBytes(MULTIPART_ENCODING));
+		out.write(byteArray);
+		String returnContent = RETURN + RETURN;
+		out.write(returnContent.getBytes(MULTIPART_ENCODING));
+	}
+	
+	private String createPartHeader(String name, String contentType, String filename) {
+        String partContent = addBoundary();
```
davidmoten: if you are reworking this use a StringBuilder.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,56 +36,158 @@ public Multipart() {
...
+        if(filename != null && name != null) {
```
davidmoten: Though the if statements may be correct you get more clarity for reading (and stop double checking values) by nesting like so:

```java
if (filename !=null) {
    if (name != null) {
        blah;
    } else {
        blah;
    }
} else if (name!=null) {
   blah;
} else {
   blah;
} 
```

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,56 +36,158 @@ public Multipart() {
...
+        	String partContentWithNameAndFilename = "Content-Disposition: form-data; name=\"%s\"; filename=\"%s\"" + RETURN + "Content-Type:%s" + RETURN + RETURN;
```
davidmoten: Are you missing a space again after ContentType:? More occurrences below.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -429,6 +434,59 @@ public void testMultipartPost(){
             fail("Unable to write to output stream");
         }
     }
+    
+    /**
+     * Test posting multipart content to a page
+     */
+    @Test
+    public void testMultipartPostWithHeadersMap(){
```
davidmoten: This is useful but what I'd like to see is coverage of the `createPartHeader` method which has a few conditional branches. If you wrote those simple tests you would see the formatting problems yourself and they would not need review iterations. Because you didn't explicilty add a test for `createPartHeader` I'm going to run through what I would expect to see for testing of a method like it:

* make `createPartHeader` static (including whatever state was used in its parameters)
* give the method default visibility
* add the annotation @VisibleForTesting or add comment // VisibleForTesting
* write numerous small test methods that cover every branch and condition of `createPartHeader` only (15 minutes of work to knock up)
* Keep the unit tests low level, only testing `createPartHeader` 

@darrelmiller I know we have discussed briefly how much unit testing to do but this is an obvious easy win and easy to write. I'm happy for testing paths not to be coverered but it should be deliberate and argued as to why it is not being done. 

On closer inspection the current test method performs no useful assertions so gives us no coverage at all of the changes made!

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -429,6 +434,59 @@ public void testMultipartPost(){
...
+            options.add(multipart.header());
+            
+            // Post the multipart content
+            OnenotePage page = orb
+            		.sections(testSection.id)
+            		.pages()
+            		.buildRequest(options)
+            		.post(multipart.content());
+            assertNotNull(page);
+        } catch (Exception e) {
```
davidmoten: try catch unnecessary, just add `throws Exception` to the method. At the moment when fails we wouldn't even get a stack trace to help us.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Used StringBuilder now for part header's value-parameter.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Restructured the conditional statements.

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
NakulSabharwal: Corrected it with space.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
NakulSabharwal: - Wrote unit tests for createPartHeader. 
- createPartHeader cannot be made static because it uses addBoundary function which uses Boundary string that can be different for different multiparts. Suppose, developer wants to use Multipart at multiple places in application at those places the boundary value will be different so it cannot be made static, in turn then addBoundary and createPartHeader cannot be made static because of usage of non-static boundary member.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
NakulSabharwal: Changed it to throw Exception.

deepak2016:
I would like to complete this PR as soon as possible, so that we can refresh SDK on Monday. @davidmoten @darrelmiller kindly review.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -429,6 +434,55 @@ public void testMultipartPost(){
...
+    	List<Option> options = new ArrayList<Option>();
+    	options.add(multipart.header());
+
+    	// Post the multipart content
+    	OnenotePage page = orb
+    			.sections(testSection.id)
+    			.pages()
+    			.buildRequest(options)
+    			.post(multipart.content());
+    	assertNotNull(page);
```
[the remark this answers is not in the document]
NakulSabharwal: The assertions should be made for checking notNull or comparison with expected url, request headers, json?

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,56 +37,160 @@ public Multipart() {
...
+	 * @param contentValue The content header value
+	 * @param contentDispParameter Map containing content paramter's key and value pair 
+	 * @return content header value and parameter string 
+	 */
+	public static String createContentHeaderValue(String contentValue, Map<String, String> contentDispParameter) {
+		String contentHeaderValue = contentValue;
+		
+		if(contentDispParameter != null) {
+			for(Map.Entry<String,String> entry : contentDispParameter.entrySet())
+	            contentHeaderValue += ";" + entry.getKey() + "=\"" + entry.getValue() + "\"";
```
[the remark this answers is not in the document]
NakulSabharwal: indented now. Although IDE was showing indented earlier.

src/test/java/com/microsoft/graph/functional/OneNoteTests.java
```diff
@@ -429,6 +437,65 @@ public void testMultipartPost(){
...
+    	assertEquals(expectedRequestUrl, pageReq.getRequestUrl());
+    	IOnenotePageCollectionRequest request = pageReq.buildRequest(options);
+    	assertNotNull(request);
+    	OnenotePageCollectionRequest pageCollectionReq = (OnenotePageCollectionRequest)request;
+    	List<HeaderOption> headeroption = pageCollectionReq.getHeaders();
+    	assertEquals("Content-Type", headeroption.get(0).getName());
+    	String expectedHeaderValue = "multipart/form-data; boundary=\""+multipart.getBoundary()+"\"";
+    	assertEquals(expectedHeaderValue, headeroption.get(0).getValue().toString());
+    	assertNotNull(multipart.content());
+    	OnenotePage page = pageReq.buildRequest(options)
```
deepak2016: you have already taken out request object in line 487, why making pageReq.buildRequest(options) call again?

NakulSabharwal: changed it to use request.post(...)

src/main/java/com/microsoft/graph/models/extensions/Multipart.java
```diff
@@ -33,56 +37,160 @@ public Multipart() {
...
+        return partContent.toString();
+	}
+	
+	/**
+	 * Create content headers value and parameter
+	 * @param contentValue The content header value
+	 * @param contentDispParameter Map containing content paramter's key and value pair 
+	 * @return content header value and parameter string 
+	 */
+	public static String createContentHeaderValue(String contentValue, Map<String, String> contentDispParameter) {
```
darrelmiller: I would just drop this method.  There isn't any commonality between Content-Type and Content-Disposition values.  We can build helper methods for creating header values that are semi-colon delimited parameter lists at a later point that can be used in other places.  This method just creates confusion and we are not using it except for in tests.

NakulSabharwal: The content header's value and parameter string building is common in Content-Type and Content-Disposition. eg. 
Content-Disposition:form-data; name="someName"
Content-Type: text/html; charset="someVal"
For this common structure of building I used one helper function for both headers.

NakulSabharwal:
@davidmoten As all major comments stand resolved, we are going ahead and merging this PR, kindly open a github issue if you see any issue with the code and we will take that as a separate PR/issue.

PR #148, opened 2018-10-01
Need option to paginate the results based on skiptoken parameter

### LINKED ISSUE #233: Skiptoken methods are not available for pagination
Hi Team,

I want to paginate the request using Skiptoken parameter. I don't see any option to do that. How can we do that with existing api? If it is not supported can you please support that?

Few interfaces reference,
IDriveItemCollectionPage
IDriveItemSearchCollectionPage

samrajcse:
fixes #233 ??
I would like to paginate the content based in skiptoken and not with pagination object. I do pagiantion on demand. So i need the option to pass the skiptoken as like top paramter. So added the option to add skiptoken in the reuest. 

Now I can do like this to paginate.

**Sample code:**

do {

				List<? extends Option> requestOptions = new ArrayList<>();
				graphClient.users(userAccount).drive().root().children("").buildRequest(requestOptions);
				IDriveItemCollectionPage rootItems = graphClient.users(userAccount).drive().root().children().buildRequest().top(5).skipToken(skipToken).get();
				List<DriveItem> root = new ArrayList<>();
				root = rootItems.getCurrentPage();
				IDriveItemCollectionRequestBuilder iDriveItemCollectionRequestBuilder = rootItems.getNextPage();
				if (iDriveItemCollectionRequestBuilder != null) {
					String nextRequestUrl = iDriveItemCollectionRequestBuilder.getRequestUrl();
					System.out.println("Next page Request url : " + nextRequestUrl);
					if (StringUtils.isNotBlank(nextRequestUrl)) {
						skipToken = StringUtils.substringAfterLast(nextRequestUrl, "skiptoken=");
					}

				} else {
					skipToken = null;
				}

				for (DriveItem driveItem : root) {
					System.out.println("##########");
					System.out.println("ID : " + driveItem.id);
					System.out.println("NAME : " + driveItem.name);
					if (driveItem.file != null) {
						System.out.println("TYPE : File");

					} else {
						System.out.println("TYPE : Folder");

					};
				}
			} while (StringUtils.isNotBlank(skipToken));


baywet:
Thank you immensely for the pull request and for your patience.
Unfortunately I can't merge your pull request as is as this section of the code is generated, but it was really helpful identifying the issue and coming up with a solution.
You can see the result of the fixed code generation here #443 
You can see the changes in the code generation process here https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/280
I also added the skip method which was missing.
Closing.

PR #163, opened 2018-11-05
Allow custom HttpURLConnection in UrlConnection used by GraphServiceClient

gazgeek:
Override constructors to allow for a custom IHttpURLConnectionProvider to be provided.
This allows for a convenient way to set the read and connect timeouts.

- Allow for a custom HttpURLConnection to be provided for GraphServiceClient. This makes it easy to set the read and connect timeouts - something that is not currently possible


deepak2016:
@gazgeek This is infra related issue and you don't need to fix anything due to this error. We are looking into this issue.

Krisoblucki-okta:
@gazgeek @deepak2016 would it make sense to allow these values to be custom set instead of being hardcoded?

NakulSabharwal:
@gazgeek This issue of configuring timeouts has been fixed in this PR https://github.com/microsoftgraph/msgraph-sdk-java/pull/216

PR #167, opened 2018-11-21
#166 - added default connect and read timeout to UrlConnection

### LINKED ISSUE #166: UrlConnection does not have neither connect timeout nor read timeout set
There is no read/connect timeouts set for UrlConnection. Because of that when any endpoint is called during which network goes down connection gets stuck (by default there is infinite timeout)

### Expected behavior
Connection times outs when network is down during a call

### Actual behavior
Connection gets stuck when network is down

while(true){
  graphClient.anyRequest();
}

Disable network during executing the loop

rgrebski:
Fixes #166

- added default connection and read timeout set to 30s

### DIFF
--- a/src/main/java/com/microsoft/graph/http/UrlConnection.java
+++ b/src/main/java/com/microsoft/graph/http/UrlConnection.java
@@ -40,6 +40,14 @@ import java.util.Map;
 public class UrlConnection implements IConnection {
 
     /**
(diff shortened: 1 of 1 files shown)


deepak2016:
Thanks for raising pull request, changes proposed makes sense to me.
@darrelmiller wanted you opinion on the numbers used.

PR #179, opened 2018-12-12
Generated models and request builders using Typewriter

MIchaelMainer:


MIchaelMainer:
The CI errors are from missing metadata on the TeamsApp, appDefinition nav property. We'll throw this PR away, and I'll see about running this with the correct metadata.

PR #190, opened 2018-12-20
Implement configurable connection timeouts on the DefaultHttpProvider

Krisoblucki-okta:
Currently there is no way to configure the connection timeouts to custom values as they are hard coded.  This change allows connection timeouts to be configured.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,9 +33,3 @@ hs_err_pid*
 /target/
 /pom.xml
 local.properties
-
-# IntelliJ
-.idea
-
-# Output
-/out
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ hs_err_pid*
 /build/
 /bin/
 
-# Eclipse
+#Eclipse
 .project
 .classpath
 .settings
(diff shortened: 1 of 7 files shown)


Krisoblucki-okta:
@deepak2016 @rgrebski 

Could you guys take a look at this and see if that makes sense.  Ideally we may want to have a IConnectionConfig object encapsulating the timeouts and maybe some other potential connection configuration settings.  What do you think?

.gitignore
```diff
@@ -33,3 +33,9 @@ hs_err_pid*
 /target/
 /pom.xml
 local.properties
+
+# IntelliJ
+.idea
+
+# Output
+/out
```
NakulSabharwal: Please revert changes to this file.

NakulSabharwal:
Hi, DefaultConnectionConfig will contain the values of timeout which can be changed using set and get. This config will be passed using getter and setter in DefaultHttpProvider to change value of IConnectionConfig connectionConfig. Timeout and connectionConfig cannot be made final as they can be changed.
The intended usage will be -
`graphClient.getHttpProvider().setConnectionConfig(any class implementing IConnectionConfig);`

deepak2016:
We have fixed the issue using PR: https://github.com/microsoftgraph/msgraph-sdk-java/pull/216
Closing this PR.

PR #197, opened 2018-12-28
Select causes 400 bad request

### LINKED ISSUE #193: specifying .select causes a 400 : Bad Request
### Expected behavior
Specifying .select(FIELDS_TO_RETURN) should return a specific set of fields on each request to post/get/patch

### Actual behavior
400 Bad Request the $select field appears twice in the request url

IUserCollectionRequest request = graphServiceClient.users().buildRequest().select(FIELDS_TO_RETURN);
            return request.post(user);

Krisoblucki-okta:
Core change to be able to fix #193 

Logic to retrieve base request url and retrieve after mutation needs to be decoupled


baywet:
thanks for your patience on the matter. I just authored https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/281 which fixes the issue by passing only the header options and you can see the resulting code changes in #445 .
This PR will be closed when the PR gets merged.

PR #203, opened 2019-01-08
Generated models and request builders using Typewriter

MIchaelMainer:


darrelmiller:
Is this PR so large because Typewriter doesn't have the most up to date templates and the extensions have been manually deleted from this repo?  /cc @NakulSabharwal @MIchaelMainer

MIchaelMainer:
We haven't yet deleted the extension files and we are still generating the extension files. This is failing due to the extension imports causing too many in memory references. I think we'll need to postpone Java automation until we get this extension story worked out.

PR #206, opened 2019-01-11
Put added in custom request.

### LINKED ISSUE #185: PUT method in CustomResquest class
Hi,

I'm trying to use CustomRequest to create a team from an already existing group (an education class, really), as there is still no teams implementation in the SDK.

The problem is that the CustomRequest class does not provide a put method. And setting the httpMethod in the GraphClient is useless, as far as I'm concerned.

I have modified the CustomRequest class including the put method, and it works fine, but it would be great if it where included in the project.

### Expected behavior

CustomRequest<JsonObject> teamRequest = graphClient.customRequest("/groups/" + educationClass.id + "/team").buildRequest();			
JsonObject jsonRequestBody...
JsonObject result = teamRequest.**put**(jsonRequestBody);

### Actual behavior

CustomRequest<JsonObject> teamRequest = graphClient.customRequest("/groups/" + educationClass.id + "/team").buildRequest();			
JsonObject jsonRequestBody...
JsonObject result = teamRequest.???(jsonRequestBody); --> No PUT method available

NakulSabharwal:
Fixes #185 

- Put functionality added in Custom request.

### DIFF
--- a/src/test/java/com/microsoft/graph/functional/CustomRequestTests.java
+++ b/src/test/java/com/microsoft/graph/functional/CustomRequestTests.java
@@ -9,7 +9,10 @@ import org.junit.Test;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.microsoft.graph.logger.DefaultLogger;
 import com.microsoft.graph.models.extensions.User;
+import com.microsoft.graph.models.extensions.UserActivity;
+import com.microsoft.graph.serializer.DefaultSerializer;
 
 /**
  * Tests for sending custom requests using the SDK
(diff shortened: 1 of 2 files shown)


src/test/java/com/microsoft/graph/functional/CustomRequestTests.java
```diff
@@ -41,4 +43,22 @@ public void testCustomGet() {
 		assertEquals(meGraphService.displayName, meOriginal.displayName);
 		assertEquals(meJson.get("displayName").getAsString(), meOriginal.displayName);
 	}
+	
+	/**
+	 * Test PUT with a custom request for both serialized and JSON content
+	 */
```
deepak2016: you added two put methods, but only one test, can you please add one more test covering other function?

NakulSabharwal: Updated tests to cover both Put functions added.

deepak2016: hey, ideally you should create two separate tests for two methods.

darrelmiller:
I'd like to discuss further why we actually have a customRequest class.  When we move to the Apache HttpClient library that has a native request object, will this be needed anymore?  The overall idea is that we can convert the output of a Request Builder to a native request object , so that users can do what ever manipulation you want.  

Adding PUT to the customRequest class feels like we are slowly re-inventing a native HTTP library.

deepak2016:
As discussed, we are going ahead and merging this pull request.

PR #266, opened 2019-07-02

MIchaelMainer:


bilalgce:
Changes already merged.

PR #274, opened 2019-07-17
Enhanced CalendarSerializer and fixed bugs in deserialization

bilalgce:
- Enhanced CalendarSerializer to deserialize dates with positive and negative offsets along in various formats
- Fixed bugs while deserializing date with millisecond and negative timezone offset
- Added unit tests for all cases of deserialization

### Other links
- https://en.wikipedia.org/wiki/ISO_8601

### DIFF
--- a/src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
+++ b/src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
@@ -55,7 +55,8 @@ public final class CalendarSerializer {
         // that can be parsed in Java
         final boolean hasZ = strVal.indexOf('Z') != -1;
         final boolean hasDot = strVal.indexOf('.') != -1;
-        
+		
+
         String modifiedStrVal;
         final String zSuffix;
         if (hasZ && hasDot) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
```diff
@@ -51,39 +51,73 @@ private CalendarSerializer() {
...
-        // Parse the well-formatted date string.
+        
+        final boolean hasOffset = modifiedStrVal.contains("T")
+				? (modifiedStrVal.substring(modifiedStrVal.indexOf('T') + 1).contains("+"))
+						|| (modifiedStrVal.substring(modifiedStrVal.indexOf('T') + 1).contains("-"))
+				: false;
+						
+        // Parse the well-formatted date string with and without offsets (eg: 2019-06-21T17:12:35.138, 2019-06-21T17:12:35.138+0000, 2019-06-21T17:12:35.138-07:00)
         final String datePattern;
         if (modifiedStrVal.contains(".")) {
```
naveenkumaragarwal: hasDot already has if date string contains dot or not. No need to compute again.

src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
```diff
@@ -51,39 +51,73 @@ private CalendarSerializer() {
...
+        	
+			String offsetSuffix = modifiedStrVal.substring(modifiedStrVal.indexOf(".") + 1);
+			
+			// Find index of offset
+			int offsetIndex = -1;
+			if (hasOffset) {
+				offsetIndex = (offsetSuffix.indexOf('+') != -1) ? offsetSuffix.indexOf('+') : offsetSuffix.indexOf('-');
+				offsetIndex = modifiedStrVal.indexOf(".") + 1 + offsetIndex; //find offset index in original string
```
naveenkumaragarwal: Use char '.' instead of string

src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
bilalgce: Done.

src/main/java/com/microsoft/graph/serializer/CalendarSerializer.java
bilalgce: Done.

PR #315, opened 2019-10-30
Update drive item

yudix:


baywet:
Hey @yudix thanks for the PR!
I'm doing a cleanup of the repo (I'm new here), trying to align branches, close issues that are already resolved and merged pull requests which waited a while.
To me it looks like most of you changes were already in master, but never merged back in dev, see #414 
Can you explain further why you added the property on the Drive item please?

yudix:
Hi,

It was indeed a long time ago, and some of my code indeed merged with
another developers PR...

Now, there is a support for conflict behavior in java
driveItem.additionalDataManager().put("@microsoft.graph.conflictBehavior

But when I send the PR the code on the doc
https://docs.microsoft.com/en-us/graph/api/driveitem-post-children?view=graph-rest-1.0&tabs=java
Was not even compile.

Now, it just a workaround and not a Java solution, I implement an enum for
this, and also there is no much info what type of conflict behavior one
should expect...
Sorry, I probably will not send more PR's, Thanks for your attention but it
is too late.
yudix.

baywet:
thanks for the quick reply and sorry it took so long to this PR to get looked at. Closing now that we have all the pieces in place (patch in the handlers, additional properties manager and docs)

PR #333, opened 2020-02-10
Fix issue #284

### LINKED ISSUE #284: Leaking connections when downloading content from OneDrive
Downloading content from OneDrive:
```
 try (InputStream req = graphClient.drives(driveId).items(docId).content().buildRequest().get()) {
            try (ByteArrayOutputStream result = new ByteArrayOutputStream()) {
//Get the content and use it - all OK
           }
}
The input and output streams are in try-with-resource blocks and so are closed at the end of the call.

In the log I then get the message:
OkHttpClient A connection to https://graph.microsoft.com/ was leaked. Did you forget to close a response body?
```

### Expected behavior

Calling close on the input stream should release the connection? Or there should be some other way to close it.

dpduong:
Fixes #284 

- I think we should close the response in finally blocks. It will help us avoid leak connections.

### DIFF
--- a/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
@@ -269,6 +269,7 @@ public class CoreHttpProvider implements IHttpProvider {
 					.tag(RetryOptions.class, retryOptions);
 			
 			String contenttype = null;
+			Response response = null;
 
 			try {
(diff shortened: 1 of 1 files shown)


rocketraman:
Any updates on this?

rocketraman:
I submitted https://github.com/microsoftgraph/msgraph-sdk-java/pull/406 which I think supersedes this one. It closes inside the finally block, and covers another potential leak as well.

PR #375, opened 2020-06-03
Fix issue #374 driveItem.additionalDataManager().put() not working

### LINKED ISSUE #374: driveItem.additionalDataManager().put("@microsoft.graph.conflictBehavior", new JsonPrimitive("rename")) not working
### Expected behavior

After enable the flag "@microsoft.graph.conflictBehavior": "rename", the duplicate folder will create with modified name (if existing folder name is "folder" then new folder will create with name "folder1")

Code snippet:
DriveItem driveItem = new DriveItem();
driveItem.name = folder;    
driveItem.folder = new Folder();
driveItem.folder.additionalDataManager().put("@microsoft.graph.conflictBehavior", new JsonPrimitive("rename"));

### Actual behavior
Giving conflict error: com.microsoft.graph.http.GraphServiceException: Error code: nameAlreadyExists
Error message: The specified item name already exists.
SdkVersion : graph-java/v1.7.1
SdkVersion : graph-java/v1.7.1
Authorization : Bearer [...]
{"folder":{},"name":"folder"}

409 : Conflict

1. Using Micosoft graph SDK 1.7.1 version
2. Call "Create folder" api two times with same folder name

vikasprabhakar:
Fixes #374 

-Code snippet:
DriveItem driveItem = new DriveItem();
driveItem.name = folder;
driveItem.folder = new Folder();
driveItem.folder.additionalDataManager().put("@microsoft.graph.conflictBehavior", new JsonPrimitive("rename"));

The additionalDataManager content will not add in request payload because of transient type.
Actual payload : {"folder":{},"name":"folder"}
Expected payload: {"folder":{},"name":"folder", "@microsoft.graph.conflictBehavior":"rename"}

In DefaultSerializer.java file, the method fieldIsOdataTransient() not considering the "@microsoft.graph.conflictBehavior" entry.
Added check for "@microsoft.graph.conflictBehavior" entry same as "@odata.type" entry.

### Other links
-

### DIFF
--- a/src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
+++ b/src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
@@ -266,7 +266,7 @@ public class DefaultSerializer implements ISerializer {
     }
 
     private boolean fieldIsOdataTransient(Map.Entry<String, JsonElement> entry) {
-        return (entry.getKey().startsWith("@") && entry.getKey() != "@odata.type");
+        return (entry.getKey().startsWith("@") && !entry.getKey().equals("@odata.type") && !entry.getKey().equals("@microsoft.graph.conflictBehavior"));
     }
     
     /**
--- a/src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
+++ b/src/main/java/com/microsoft/graph/serializer/DefaultSerializer.java
@@ -256,17 +256,8 @@ public class DefaultSerializer implements ISerializer {
      */
     private void addAdditionalDataToJson(AdditionalDataManager additionalDataManager, JsonObject jsonNode) {
     	for (Map.Entry<String, JsonElement> entry : additionalDataManager.entrySet()) {
-            if (!fieldIsOdataTransient(entry)) {
-                jsonNode.add(
-                        entry.getKey(),
-                        entry.getValue()
-                );
-            }
+                jsonNode.add(entry.getKey(), entry.getValue());
         }
-    }
-
-    private boolean fieldIsOdataTransient(Map.Entry<String, JsonElement> entry) {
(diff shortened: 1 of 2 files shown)


ddyett:
@MIchaelMainer do you know of any other types we should cover here or is this good for now.  i think you mentioned we don't have this check in dotnet at all.

vikasprabhakar:
@MIchaelMainer, request you review the changes.

MIchaelMainer:
I want to get rid of this transient property check. @vikasprabhakar can you update your PR so that the transient property check is removed? 

This will result in all properties instance annotations (transient properties) in the additionalPropertyManager being sent over the wire. The service API is *supposed* to gracefully ignore instance annotations it doesn't expect. The client MUST be able to set any instance annotation.

This exists because developers incorrectly use our deserialized models for their models, instead of using our models as a state representation. So, a model is received using GET, the client updates that model, and then uses it in a subsequent PATCH, thus sending not only transient properties, but also read-only properties from the original GET. This transient property check addressed the scenario I describe but blocked other potential scenarios, like the one in this PR.

Filtering out transient property with an allowlist in the client library is not ideal since transient properties can be API requirements at the service API without notice to the client.

cc: @ramsessanchez @ddyett

vikasprabhakar:
Thanks @https://github.com/MIchaelMainer for merge the changes.
I am new in this forum and not aware of SDK release process. May I know when the next release plan with the changes?
Thank you again.

MIchaelMainer:
You're welcome @vikasprabhakar. Thank you for investigating and making this change. Much appreciated. We released this today. Update your dependency to 1.8.0 and let us know.

vikasprabhakar:
Thank you @MIchaelMainer for releasing the jar 1.8.0 with changes. I checked with 1.8.0 jar and it is working as expected.

PR #397, opened 2020-07-15
[#396] OneDrive Chunked Upload - Chunk size is not taken into consideration in case of customized input stream

### LINKED ISSUE #396: OneDrive Chunked Upload - Chunk size is not taken into consideration in case of customized input stream
### Expected behavior
Large files should be uploaded in chunks using the specified chunk size according to the following Javadoc comment of the upload method in **ChunkedUploadProvider** class.
/**
     * Uploads content to remote upload session based on the input stream
     *
     * @param options  the upload options
     * @param callback the progress callback invoked during uploading
     * @param **_configs  the optional configurations for the upload options. [0] should be the customized chunk_**
     *                 **_size_** and [1] should be the maxRetry for upload retry.
     * @throws IOException the IO exception that occurred during upload
     */
    ChunkedUploadProvider.upload(final List<Option> options,
                       final IProgressCallback<UploadType> callback,
                       final int... configs)

### Actual behavior
In enterprise product environment with high parallelism, input streams are often controlled on global level setting data to be read in smaller chunks, like 32 KB. For example the **InputStream read(byte[] buffer)** method reads implicitly less than the specified buffer/chunk size (e.g. reads 32 KB while chunk size is 5 MB). Currently the chunks that are being uploaded are with size equal to what is read from the stream - 32 KB. This behavior leads to too many requests in case of large files.

1. Use the sample code described [here](https://docs.microsoft.com/en-us/graph/sdks/large-file-upload?tabs=java) but specify 5 MB chunk size.
2. Use following custom stream in the sample:
`InputStream fileStream = new MyInputStream(new FileInputStream(file))`
instead of:
`InputStream fileStream = new FileInputStream(file);`

public class MyInputStream extends InputStream {

    private InputStream sourceIs;

    public MyInputStream(InputStream sourceIs) {
        this.sourceIs = sourceIs;
    }

    public int read() throws IOException {
        return sourceIs.read();
    }

    @Override
    public int read(byte[] b) throws IOException {
        return super.read(b, 0, 32 * 1024);
    }
}
3. Run the sample code.

**Sample output:**
Connecting...
Uploaded 0 bytes of 5.296497344970703 MB.
Uploaded 32.0 KB of 5.296497344970703 MB.
Uploaded 64.0 KB of 5.296497344970703 MB.
Uploaded 96.0 KB of 5.296497344970703 MB.
...

**Client library versions used are:**
- microsoft-graph 1.8.0
- microsoft-graph-core 1.0.1
- microsoft-graph-auth 0.2.0

Pull request: https://github.com/microsoftgraph/msgraph-sdk-java/pull/397

katestoycheva:
Fixes #396 

- Continue to read from the input stream to make sure the buffer is filled with data until the chunk size is reached before executing the chunked upload request.

### DIFF
--- a/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
+++ b/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
@@ -168,10 +168,10 @@ public class ChunkedUploadProvider<UploadType> {
         byte[] buffer = new byte[chunkSize];
 
         while (this.readSoFar < this.streamSize) {
-            int read = 0;
             int buffRead = 0;
 
             while (buffRead < chunkSize) {
+                int read = 0;
                 read = this.inputStream.read(buffer, buffRead, chunkSize - buffRead);
                 if (read == -1) {
                     break;
--- a/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
+++ b/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
@@ -168,14 +168,19 @@ public class ChunkedUploadProvider<UploadType> {
         byte[] buffer = new byte[chunkSize];
 
         while (this.readSoFar < this.streamSize) {
-            int read = this.inputStream.read(buffer);
+            int read = 0;
+            int buffRead = 0;
 
-            if (read == -1) {
-                break;
+            while (buffRead < chunkSize) {
+                read = this.inputStream.read(buffer, buffRead, chunkSize - buffRead);
+                if (read == -1) {
+                    break;
+                }
+                buffRead += read;
             }
 
             ChunkedUploadRequest request =
-                    new ChunkedUploadRequest(this.uploadUrl, this.client, options, buffer, read,
+                    new ChunkedUploadRequest(this.uploadUrl, this.client, options, buffer, buffRead,
                             maxRetry, this.readSoFar, this.streamSize);
             ChunkedUploadResult<UploadType> result = request.upload(this.responseHandler);
 
@@ -190,7 +195,7 @@ public class ChunkedUploadProvider<UploadType> {
                 break;
             }
 
-            this.readSoFar += read;
+            this.readSoFar += buffRead;
         }
     }
     
--- a/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
+++ b/src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
@@ -170,6 +170,7 @@ public class ChunkedUploadProvider<UploadType> {
         while (this.readSoFar < this.streamSize) {
             int buffRead = 0;
 
+            // inner loop is to work-around the case where read buffer size is limited to less than chunk size by a global setting
             while (buffRead < chunkSize) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
```diff
@@ -168,14 +168,19 @@ public void upload(final List<Option> options,
         byte[] buffer = new byte[chunkSize];
 
         while (this.readSoFar < this.streamSize) {
-            int read = this.inputStream.read(buffer);
-
-            if (read == -1) {
-                break;
+            int read = 0;
```
zengin, line 171: We should make this local to the inner while loop to scope the variable better.

src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
```diff
@@ -168,14 +168,19 @@ public void upload(final List<Option> options,
...
+                read = this.inputStream.read(buffer, buffRead, chunkSize - buffRead);
+                if (read == -1) {
+                    break;
+                }
+                buffRead += read;
             }
 
             ChunkedUploadRequest request =
-                    new ChunkedUploadRequest(this.uploadUrl, this.client, options, buffer, read,
+                    new ChunkedUploadRequest(this.uploadUrl, this.client, options, buffer, buffRead,
```
zengin, line 183: By looking at the local scope, it seems possible to reach here with `buffRead == 0`, if we break out of the inner while loop with `read == -1`. Previous code seems to have been checking for that.

That being said, maybe previous code was over-defensive given that reaching end of stream (`read == -1`) would have also meant the `while` condition `(this.readSoFar < this.streamSize)` is false.

katestoycheva:
Hi @zengin 
I re-scoped the read variable and I confirm that your interpretation about not checking the empty buffer is correct.
Please review and let me know if anything else needs to be changed.

src/main/java/com/microsoft/graph/concurrency/ChunkedUploadProvider.java
```diff
@@ -168,14 +168,19 @@ public void upload(final List<Option> options,
...
 
         while (this.readSoFar < this.streamSize) {
-            int read = this.inputStream.read(buffer);
-
-            if (read == -1) {
-                break;
+            int buffRead = 0;
+
+            while (buffRead < chunkSize) {
+                int read = 0;
```
zengin, line 174:
```suggestion
            // inner loop is to work-around the case where read buffer size is limited to less than chunk size by a global setting
            while (buffRead < chunkSize) {
                int read = 0;
```

davidmoten:
@katestoycheva just for curiosity how significant is this? What is this custom input stream?  For my scenarios I don't care if a chunk ends up half the requested size but I do care if the chunk ends up 1% of the requested size repeatedly.

zengin:
@davidmoten, there are more details in the attached issue. I think in @katestoycheva's example, the ratio is 32KB/5MB = 1/160.

https://github.com/microsoftgraph/msgraph-sdk-java/issues/396

davidmoten:
@zengin thanks!

Issue #313, opened 2019-10-24
ClientException thrown when tcp connection is closed


James-Ronayne:
Expected behavior
When a underlying connection okhttp is using is closed okhttp should reopen a new connection without throwing an exception.
Actual behavior
ClientException is thrown by the graph sdk.
Underlying connection okhttp is handling is closed causing a EOFException to be thrown. This eventually bubbles up to the sdk library and is caught and rethrown as a ClientException.
Steps to reproduce the behavior
Make a number of calls to the graph sdk consistantly over a period of 15-30 minutes and the error should be seen.
This looks to be caused by setting okBuilder.retryOnConnectionFailure(false); in CoreHttpProvider. I can't see why this was done but it happened when RetryHandler functionality was added from msgraph-sdk-java-core. Since RetryHandler only retries status code responses of 429 503 504 and doesn't actually do anything about exeptions this should be set to true.
There's no way to override this setting the only workaround I've found is to add retry logic into any code that makes a call with the sdk or inject a 'Connection: close' header into every request made making okhttp open a new connection everytime.
relevant info: square/okhttp/issues/1873
com.microsoft.graph.core.ClientException: Error during http request
com.microsoft.graph.http.CoreHttpProvider.sendRequestInternal(CoreHttpProvider.java:422) ~[microsoft-graph-1.6.0.jar!/:na]\tat com.microsoft.graph.http.CoreHttpProvider.send(CoreHttpProvider.java:204) ~[microsoft-graph-1.6.0.jar!/:na]
 com.microsoft.graph.http.CoreHttpProvider.send(CoreHttpProvider.java:184) ~[microsoft-graph-1.6.0.jar!/:na]
 com.microsoft.graph.http.BaseCollectionRequest.send(BaseCollectionRequest.java:89) ~[microsoft-graph-1.6.0.jar!/:na]
 com.microsoft.graph.requests.extensions.UserCollectionRequest.get(UserCollectionRequest.java:52) ~[microsoft-graph-1.6.0.jar!/:na]
.......
 org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61) [tomcat-embed-core-9.0.24.jar!/:9.0.24]
 java.lang.Thread.run(Thread.java:748) [na:1.8.0_222]
Caused by: java.io.IOException: unexpected end of stream on https://graph.microsoft.com/...
[...]

PR #406, opened 2020-08-06
Fix some OkHttp response body leaks

### LINKED ISSUE #284: Leaking connections when downloading content from OneDrive

rocketraman:
Fixes #284 
Closes #333

- OkHttp response bodies are required to be closed. See https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/. There are a couple of places in which this does not happen, and this pull should fix at least some of those leaks.

The first is in `CoreHttpProvider`. In the cases in which a `byteStream` is obtained from the response body, the stream is closed, which closes the body, so those are ok. However, in some other cases, the Response is used without closing it. Now, closing the Response in the finally method covers all those situations.

The second is in `GraphServiceException` which reads the response body but did not close it.

### Other links
- https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/

### DIFF
--- a/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
@@ -269,6 +269,7 @@ public class CoreHttpProvider implements IHttpProvider {
 					.tag(RetryOptions.class, retryOptions);
 			
 			String contenttype = null;
+			Response response = null;
 
 			try {
 				logger.logDebug("Request Method " + request.getHttpMethod().toString());
@@ -353,7 +354,7 @@ public class CoreHttpProvider implements IHttpProvider {
 				coreHttpRequest = corehttpRequestBuilder.build();
 
 				// Call being executed
-				Response response = corehttpClient.newCall(coreHttpRequest).execute();
+				response = corehttpClient.newCall(coreHttpRequest).execute();
 
 				if (handler != null) {
 					handler.configConnection(response);
@@ -401,12 +402,13 @@ public class CoreHttpProvider implements IHttpProvider {
 					return (Result) handleBinaryStream(in);
 				}
 			} finally {
-				if (!isBinaryStreamInput && in != null) {
+				if (!isBinaryStreamInput) {
 					try{
-						in.close();
+						if (in != null) in.close();
 					}catch(IOException e) {
 						logger.logError(e.getMessage(), e);
 					}
+					if (response != null) response.close();
 				}
 			}
 		} catch (final GraphServiceException ex) {
--- a/src/main/java/com/microsoft/graph/http/GraphServiceException.java
+++ b/src/main/java/com/microsoft/graph/http/GraphServiceException.java
@@ -23,6 +23,7 @@
 package com.microsoft.graph.http;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
@@ -38,6 +39,7 @@ import com.microsoft.graph.options.HeaderOption;
 import com.microsoft.graph.serializer.ISerializer;
 
 import okhttp3.Response;
+import static okhttp3.internal.Util.closeQuietly;
 
 /**
  * An exception from the Graph service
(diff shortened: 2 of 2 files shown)


rocketraman:
Have been running this for a day or so and so far I've not noticed any further leak reports from OkHttp.

yaelal3:
this is an important fix, plz merge it and release a patch, as it causes crashes when using the api

ravisamani:
> Have been running this for a day or so and so far I've not noticed any further leak reports from OkHttp.

good!

rocketraman:
@ravisamani Yes, it is different, *however* I have a theory that could possibly related them in another way: if many people are running SDKs with leaking connection code (and they are), then its possible the graph server could be overloaded with stale connections, and occasionally and randomly resets active connections as a way to avoid a wider DOS. That might then result in the connection reset errors people are seeing in issue #313 . Though that's just a wild-ass guess because I have no insight or visibility into Microsoft's backend.

ravisamani:
we'll need upgrade as we still continue to get this errors.

any release date of 1.10.0 ?

ravisamani:
> we still continue to get - **Connection Reset** errors - after changing 2 things at our end:
> 
> 1. reuse IGraphServiceClient instance for all requests made - **Singleton**
> 2. Timeout increased from 30s to 60s
> 
> looking forward to 1.10.0 release - August 28, 2020

@baywet  we've doubt for #point - 1 above ?  

what should be ideal life cycle for GraphServiceClient ?

For Each Graph API Request : 
1 -  create new GraphServiceClient 
2 -  reuse singleton GraphServiceClient

baywet:
Hey @ravisamani 
Are you having those issues with version 2.0.0 or an earlier version?
The general practice is to keep the graph client for a given authentication context:
- if you're building a backend service relying on app only context, you can keep it for a long time/use a singleton
- if you're building a backend API/webapp that acts on behalf of the user, keep the client for the duration of the user's "session"
- if you're building a native delegated app (android for example), keep the client as long as the app is active and the user signed in

ravisamani:
> Hey @ravisamani
> Are you having those issues with version 2.0.0 or an earlier version?
> The general practice is to keep the graph client for a given authentication context:
> 
> * if you're building a backend service relying on app only context, you can keep it for a long time/use a singleton
> * if you're building a backend API/webapp that acts on behalf of the user, keep the client for the duration of the user's "session"
> * if you're building a native delegated app (android for example), keep the client as long as the app is active and the user signed in

We have those issues with **1.7.1** . 

We used singleton GraphServiceClient for all users - ( And made sure that IAuthenticationProvider picks correct user's AccessToken )

baywet:
Can you try with 2.0.0 please since this pull request was release with 2.0.0?

As you pointed out, it really depends on the implementation of the Authentication provider and the execution context. Some people might end up implementing authentication providers that are "sticky" and in some context send a request with an access token for user B when they meant to act on behalf of user A. If you are certain your implementation of authentication provider is correct for your context, there's no issue making the client a singleton.

ravisamani:
> Can you try with 2.0.0 please since this pull request was release with 2.0.0?
> 
> As you pointed out, it really depends on the implementation of the Authentication provider and the execution context. Some people might end up implementing authentication providers that are "sticky" and in some context send a request with an access token for user B when they meant to act on behalf of user A. If you are certain your implementation of authentication provider is correct for your context, there's no issue making the client a singleton.

Yes - we made sure at RunTime to give correct accessToken to IAuthenticationProvider. Will let you know after upgrade. 
Thanks - **Vincent**

baywet:
For that specific issue I don't think so. But I'd recommend upgrading core to 1.0.2, there are a number of issues that have been fixed and some dependencies have been removed.

ravisamani:
> For that specific issue I don't think so. But I'd recommend upgrading core to 1.0.2, there are a number of issues that have been fixed and some dependencies have been removed.

I see, which dependencies are removed?

baywet:
https://github.com/microsoftgraph/msgraph-sdk-java-core/compare/1.0.1...1.0.2
Guava, because it wasn't in use in core. (and we're bumping it for this library)

PR #412, opened 2020-08-18
Travis ci

baywet:
cleaning up old branches. I think I can safely delete this one but I wanted a confirmation before I do so


.travis.yml
```diff
@@ -1 +1,6 @@
 language: java
+jdk:
+  - openjdk7
```
baywet, line 3:
```suggestion
  - openjdk8
```

PR #467, opened 2020-09-14
- fixes a bug where the content type of a post request would be forcibly overridden

### LINKED ISSUE #357: Empty POST request rewrites content type
Hello,
I'm trying to make a call to microsoft graph API in order to revoke signin sessions.
I built the graph client like this and it is working like a charm for multiple call.
  @Bean
  fun authenticationProvider(): IAuthenticationProvider =
          OAuth2AuthenticationProvider(createResourceDetails())

  @Bean
  fun graphClient(provider: IAuthenticationProvider): IGraphServiceClient =
          GraphServiceClient
                  .builder()
                  .authenticationProvider(provider)
                  .buildClient()

  private fun createResourceDetails() =
          ClientCredentialsResourceDetails().apply {
            clientId = configuration.clientId
            clientSecret = configuration.clientSecret
            accessTokenUri = configuration.accessTokenUri
            clientAuthenticationScheme = AuthenticationScheme.form
            scope = configuration.scopes
          }

I am now trying to make the revoke signin sessions call. Here is the snippet:
  private fun revokeUserToken(userId: String) {
    try {
      val jsonContentType: Option = HeaderOption("Content-Type", MediaType.APPLICATION_JSON_VALUE)
      graphClient.users(userId).revokeSignInSessions().buildRequest(mutableListOf(jsonContentType)).post()
    } catch (e: GraphServiceException) {
      LOG.error("Cannot revoke tokens for user $userId. ${e.serviceError.message}")
    }
  }

This is Kotlin code as close to Java one but do not hesitate to ask if anything is unclear.
Concretely I make the post request using the revokeSignInSessions request and I add manually the json content type.
However microsoft grph api always fails with this issue:
Throwable detail: com.microsoft.graph.http.GraphServiceException: Error code: Request_BadRequest
Error message: A supported MIME type could not be found that matches the content type of the response. None of the supported type(s) 'application/json;odata=minimalmetadata;streaming=true, application/json;odata=minimalmetadata;streaming=false, application/json;odata=minimalmetadata, application/json;odata=fullmetadata;streaming=true, application/json;odata=fullmetadata;streaming=false, application/json;odata=fullmetadata, application/json;odata=nometadata;streaming=true, application/json;odata=nometadata;streaming=false, application/json;odata=nometadata, application/json;streaming=true, application/json;streaming=false, application/json;odata=verbose, application/json' matches the content type 'application/octet-stream'.

After looking into the code, I end up in the CoreHttpProvider in the sendRequestInternal method.
And in this methods it states that if the body is null, then it rewrites content type to octetstream (beginning is line 286 of the class)
				if (serializable == null) {
					// Send an empty body through with a POST request
					// This ensures that the Content-Length header is properly set
					if (request.getHttpMethod() == HttpMethod.POST) {
						bytesToWrite = new byte[0];
						contenttype = Constants.BINARY_CONTENT_TYPE;
					}
					else {
						bytesToWrite = null;
					}
				}

From the UserRevokeSignInSessionsRequest class, I can only make post request that have no body (begin is line 36)
    /**
     * Creates the UserRevokeSignInSessions
     *
     * @param callback the callback to be called after success or failure
     */
    public void post(final ICallback<Boolean> callback) {
        send(HttpMethod.POST, callback, null);
    }

    /**
     * Creates the UserRevokeSignInSessions
     *
     * @return the Boolean
     * @throws ClientException an exception occurs if there was an error while the request was sent
     */
    public Boolean post() throws ClientException {
       return send(HttpMethod.POST, null);
    }

so I will always end up with a octet stream content type, while this API accepts only json content type.
Is there any way to bypass this content type rewritting so that I can use all the classes you made for revoking sessions ?
Thank you
[AB#6033](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6033)

baywet:
Fixes #357 

Warning: this pull request contains #463 to ease up the writing of the unit test. 463 should be reviewed and merged first.

-

### Other links

-

### DIFF
--- a/src/test/java/com/microsoft/graph/functional/UserTests.java
+++ b/src/test/java/com/microsoft/graph/functional/UserTests.java
@@ -178,10 +178,10 @@ public class UserTests {
         final ArrayList<Option> options = new ArrayList<>();
         options.add(ctype);
         final Request request = graphServiceClient.me()
-                                            .revokeSignInSessions().
-                                            buildRequest(options)
+                                            .revokeSignInSessions()
+                                            .buildRequest(options)
                                             .withHttpMethod(HttpMethod.POST)
-											.getHttpRequest();
+                                            .getHttpRequest();
 		assertEquals(contentTypeValue, request.body().contentType().toString());					
 	}
 
--- a/src/test/java/com/microsoft/graph/functional/UserTests.java
+++ b/src/test/java/com/microsoft/graph/functional/UserTests.java
@@ -1,21 +1,28 @@
 package com.microsoft.graph.functional;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
+import okhttp3.Request;
+
+import com.microsoft.graph.http.HttpMethod;
 import com.microsoft.graph.models.extensions.Drive;
 import com.microsoft.graph.models.extensions.DriveItem;
 import com.microsoft.graph.models.extensions.IGraphServiceClient;
 import com.microsoft.graph.models.extensions.ProfilePhoto;
 import com.microsoft.graph.models.extensions.User;
+import com.microsoft.graph.options.HeaderOption;
+import com.microsoft.graph.options.Option;
 import com.microsoft.graph.requests.extensions.IContactCollectionPage;
 import com.microsoft.graph.requests.extensions.IDirectoryObjectCollectionWithReferencesPage;
 import com.microsoft.graph.requests.extensions.IDriveItemCollectionPage;
@@ -162,6 +169,20 @@ public class UserTests {
 	public void meMemberof() {
 		IDirectoryObjectCollectionWithReferencesPage page = graphServiceClient.me().memberOf().buildRequest().get();
 		assertNotNull(page);
+	}
+
+	@Test
+	public void emptyPostContentType() {
+		final String contentTypeValue = "application/json";
+		final HeaderOption ctype = new HeaderOption("Content-Type", contentTypeValue);
+        final ArrayList<Option> options = new ArrayList<>();
+        options.add(ctype);
+        final Request request = graphServiceClient.me()
+                                            .revokeSignInSessions().
+                                            buildRequest(options)
+                                            .withHttpMethod(HttpMethod.POST)
+											.getHttpRequest();
+		assertEquals(contentTypeValue, request.body().contentType().toString());					
 	}
 
 }
--- a/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
@@ -260,7 +260,9 @@ public class CoreHttpProvider implements IHttpProvider {
 			// This ensures that the Content-Length header is properly set
 			if (request.getHttpMethod() == HttpMethod.POST) {
 				bytesToWrite = new byte[0];
-				contenttype = Constants.BINARY_CONTENT_TYPE;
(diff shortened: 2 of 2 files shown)


zengin:
@baywet, is this the only commit that requires review here: https://github.com/microsoftgraph/msgraph-sdk-java/pull/467/commits/73a632b04cffc6553aeed8a6b9a13eddabf27424?

baywet:
@zengin yes, and sorry for the mess, I needed the other PR to build my unit tests

zengin:
@baywet, in these cases, rebasing the branch could help clear the diff. If that doesn't do the job, creating a new feature branch and cherry-picking relevant commits over can simplify the whole process. Then existing PR can be closed with a link to the new one.

Doing this has two benefits:
1. Short term: saving time in code reviews, especially when there are multiple repositories and cost of context switching is high.
2. Long term: revisiting old PRs, whether it is through "git bisect" or just a casual linear look over the merged PRs, will be cheap as well because the actual change is a much smaller subset. In this particular PR, we have the history due to our notes to the relevant commits, but if the reviewer knows it by other means (e.g. the discussion happened elsewhere), the context will be lost from official Github history.

Of course, everything has a trade-off and it may not be feasible to do that in certain situations. But I think, in general, the benefits for reviewers and everybody who looks at the history at a later time, including the author, outweigh the trouble of moving things around.

My two cents. Please let me know what you think.

baywet:
@zengin makes perfect sense! I'm always hesitant to force push and/or rebase in repos where it's not a common practice. It usually confuses people. But if the team is ok with that I'll use it as common practice.
I just rebased the branch which should make the review much easier :)

zengin:
@baywet I agree that force pushes shouldn't be common practice, especially on the main branches. But throw-away feature branches should be OK to be rebased. I see no downside to rebasing if the PR hasn't been created yet or reviews hasn't started. The latter one is hard to check because once the PR is created, the review may get started without any indication in the Github interface :)

src/test/java/com/microsoft/graph/functional/UserTests.java
```diff
@@ -164,4 +171,18 @@ public void meMemberof() {
...
+	public void emptyPostContentType() {
+		final String contentTypeValue = "application/json";
+		final HeaderOption ctype = new HeaderOption("Content-Type", contentTypeValue);
+        final ArrayList<Option> options = new ArrayList<>();
+        options.add(ctype);
+        final Request request = graphServiceClient.me()
+                                            .revokeSignInSessions().
+                                            buildRequest(options)
+                                            .withHttpMethod(HttpMethod.POST)
+											.getHttpRequest();
```
zengin, line 184: nit: indentation looks off.

zengin (approved):

src/test/java/com/microsoft/graph/functional/UserTests.java
```diff
@@ -164,4 +171,18 @@ public void meMemberof() {
...
 	}
 
+	@Test
+	public void emptyPostContentType() {
+		final String contentTypeValue = "application/json";
+		final HeaderOption ctype = new HeaderOption("Content-Type", contentTypeValue);
+        final ArrayList<Option> options = new ArrayList<>();
+        options.add(ctype);
+        final Request request = graphServiceClient.me()
+                                            .revokeSignInSessions().
```
baywet, line 181:
```suggestion
                                            .revokeSignInSessions()
```

src/test/java/com/microsoft/graph/functional/UserTests.java
```diff
@@ -164,4 +171,18 @@ public void meMemberof() {
...
+                                            buildRequest(options)
```
baywet, line 182:
```suggestion
                                            .buildRequest(options)
```

src/test/java/com/microsoft/graph/functional/UserTests.java
baywet:
```suggestion
                                            .getHttpRequest();
```

zengin (approved):

PR #470, opened 2020-09-15
- fixes a bug where downloading json files from onedrive would cause a cast exception

### LINKED ISSUE #275: Error when trying to download .JSON files from OneDrive
I am trying to download a .json file from OneDrive. The following request patterns get the intended behavior (the file is downloaded) for other file types but none work for .json files.
InputStream driveItemInputStream = getGraphServiceClient().drives(driveId).items(driveItemId).content().buildRequest().get();
InputStream driveItemInputStream = getGraphServiceClient().drives().byId(driveId).items().byId(driveItemId).content().buildRequest().get();
InputStream driveItemInputStream = getGraphServiceClient().customRequest(("/drives/"+driveId+"/items/"+driveItemId+"/content"), InputStream.class).buildRequest().get();

Expected behavior
For the file to be downloaded.
Actual behavior
Error response:
com.microsoft.graph.core.ClientException: Error during http request
at com.microsoft.graph.http.DefaultHttpProvider.sendRequestInternal(DefaultHttpProvider.java:368)
	at com.microsoft.graph.http.DefaultHttpProvider.send(DefaultHttpProvider.java:204)
	at com.microsoft.graph.http.DefaultHttpProvider.send(DefaultHttpProvider.java:184)
	at com.microsoft.graph.http.BaseStreamRequest.send(BaseStreamRequest.java:81)
	at com.microsoft.graph.requests.extensions.DriveItemStreamRequest.get(DriveItemStreamRequest.java:55)
...
Caused by: java.lang.RuntimeException: Failed to invoke public java.io.InputStream() with no args
	at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:113)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:212)
	at com.google.gson.Gson.fromJson(Gson.java:927)
	at com.google.gson.Gson.fromJson(Gson.java:892)
	at com.google.gson.Gson.fromJson(Gson.java:841)
	at com.google.gson.Gson.fromJson(Gson.java:813)
	at com.microsoft.graph.serializer.DefaultSerializer.deserializeObject(DefaultSerializer.java:78)
	at com.microsoft.graph.http.DefaultHttpProvider.handleJsonResponse(DefaultHttpProvider.java:417)
	at com.microsoft.graph.http.DefaultHttpProvider.sendRequestInternal(DefaultHttpProvider.java:342)
	... 46 more

Steps to reproduce the behavior
Use the SDK to make a GET request on a .json file in a user or group OneDrive using any of the request patterns above.
[AB#6032](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6032)

baywet:
Fixes #275

-

### Other links

-

### DIFF
--- a/src/main/java/com/microsoft/graph/core/Constants.java
+++ b/src/main/java/com/microsoft/graph/core/Constants.java
@@ -18,10 +18,6 @@ public final class Constants {
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
 	/**
-	 * The content type for text responses
-	 */
-	public static final String TEXT_CONTENT_TYPE = "text/plain";
-	/**
 	 * The binary content type header's value 
 	 */
 	public static final String BINARY_CONTENT_TYPE = "application/octet-stream";
--- a/src/main/java/com/microsoft/graph/core/Constants.java
+++ b/src/main/java/com/microsoft/graph/core/Constants.java
@@ -18,6 +18,10 @@ public final class Constants {
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
 	/**
+	 * The content type for text responses
+	 */
+	public static final String TEXT_CONTENT_TYPE = "text/plain";
+	/**
 	 * The binary content type header's value 
 	 */
 	public static final String BINARY_CONTENT_TYPE = "application/octet-stream";
--- a/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
@@ -417,20 +417,20 @@ public class CoreHttpProvider implements IHttpProvider {
 
 				final Map<String, String> headers = CoreHttpProvider.getResponseHeadersAsMapStringString(response);
 
+				if(response.body() == null || response.body().contentLength() == 0)
+					return (Result) null;
+
 				final String contentType = headers.get(Constants.CONTENT_TYPE_HEADER_NAME);
-				if (contentType != null && contentType.contains(Constants.JSON_CONTENT_TYPE)) {
+				if (contentType != null && resultClass != InputStream.class && 
+							(contentType.contains(Constants.JSON_CONTENT_TYPE) || contentType.contains(Constants.TEXT_CONTENT_TYPE))) { // some services reply in text/plain with a JSON representation...
 					logger.logDebug("Response json");
 					return handleJsonResponse(in, CoreHttpProvider.getResponseHeadersAsMapOfStringList(response), resultClass);
-				} else {
+				} else if (resultClass == InputStream.class) {
 					logger.logDebug("Response binary");
 					isBinaryStreamInput = true;
-					if (resultClass == InputStream.class) {
-						return (Result) handleBinaryStream(in);
-					} else if(response.body() != null && response.body().contentLength() > 0) { // some services reply in text/plain with a JSON representation...
-						return handleJsonResponse(in, CoreHttpProvider.getResponseHeadersAsMapOfStringList(response), resultClass);
-					} else {
-						return (Result) null;
-					}
+					return (Result) handleBinaryStream(in);
+				} else {
+					return (Result) null;
 				}
 			} finally {
 				if (!isBinaryStreamInput) {
--- a/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
+++ b/src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
@@ -422,7 +422,7 @@ public class CoreHttpProvider implements IHttpProvider {
 
 				final String contentType = headers.get(Constants.CONTENT_TYPE_HEADER_NAME);
 				if (contentType != null && resultClass != InputStream.class && 
-							(contentType.contains(Constants.JSON_CONTENT_TYPE) || contentType.contains(Constants.TEXT_CONTENT_TYPE))) { // some services reply in text/plain with a JSON representation...
+							contentType.contains(Constants.JSON_CONTENT_TYPE)) {
 					logger.logDebug("Response json");
 					return handleJsonResponse(in, CoreHttpProvider.getResponseHeadersAsMapOfStringList(response), resultClass);
 				} else if (resultClass == InputStream.class) {
--- a/src/test/java/com/microsoft/graph/functional/OneDriveTests.java
+++ b/src/test/java/com/microsoft/graph/functional/OneDriveTests.java
@@ -1,6 +1,7 @@
 package com.microsoft.graph.functional;
 
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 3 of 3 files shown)


src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
```diff
@@ -417,20 +417,20 @@ public Request authenticateRequest(Request request) {
 
 				final Map<String, String> headers = CoreHttpProvider.getResponseHeadersAsMapStringString(response);
 
+				if(response.body() == null || response.body().contentLength() == 0)
```
zengin, line 420: minor perf: if `body()` is more than a simple property access, consider saving a reference to that into a variable to avoid calling it twice. If it is property access, the overhead should be minimal, so you can ignore this comment in that case.

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
```diff
@@ -417,20 +417,20 @@ public Request authenticateRequest(Request request) {
...
+					return (Result) null;
+
 				final String contentType = headers.get(Constants.CONTENT_TYPE_HEADER_NAME);
-				if (contentType != null && contentType.contains(Constants.JSON_CONTENT_TYPE)) {
+				if (contentType != null && resultClass != InputStream.class && 
+							(contentType.contains(Constants.JSON_CONTENT_TYPE) || contentType.contains(Constants.TEXT_CONTENT_TYPE))) { // some services reply in text/plain with a JSON representation...
```
zengin, line 425: interesting, is there any proper non-JSON `text/plain` response from the services?

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
baywet: thanks for double checking. I had checking for that and it's simply an accessor to a private field.

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
baywet: file upload on a $value sometimes replies with a representation of the json object but in text plain (I don't remember which one right now though)

baywet: and re-reading the question I don't think I'm answering the question. I haven't seen a text/plain response with actual text. But it might be possible. Uploadprofile picture replies with 200 ok of type text/plain but the body is empty

zengin: yeah, the concern is whether we ever end up trying to parse a non-JSON text/plain response as JSON.

baywet: would you remove it or then or fail gracefully?

zengin: I found a case where it can be `text/plain` without being a JSON response. If an attachment is a plain text file, `https://graph.microsoft.com/v1.0/me/messages/<message_id>/attachments/<attachment_id>/$value` returns a `text/plain` response. That will be a problem when we attempt to parse a text file as JSON.

I think the solution should be to have the service team to return correct `Content-Type` and us to have another `else if` block for handling plain text.

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
```diff
@@ -417,20 +417,20 @@ public Request authenticateRequest(Request request) {
...
 					logger.logDebug("Response binary");
 					isBinaryStreamInput = true;
-					if (resultClass == InputStream.class) {
-						return (Result) handleBinaryStream(in);
-					} else if(response.body() != null && response.body().contentLength() > 0) { // some services reply in text/plain with a JSON representation...
-						return handleJsonResponse(in, CoreHttpProvider.getResponseHeadersAsMapOfStringList(response), resultClass);
-					} else {
-						return (Result) null;
-					}
+					return (Result) handleBinaryStream(in);
```
nikithauc, line 431: The sendRequestInternal() function is a long function. I would recommend breaking them into smaller parts. Example -> introduced a smaller function to handle different response types and call that in sendRequestInternal()

src/test/java/com/microsoft/graph/functional/OneDriveTests.java
```diff
@@ -85,4 +87,18 @@ public void testDownloadWithCustomRequest() throws IOException {
 		   assertFalse("stream should not be empty", stream.read() == -1);
 		}
 	}
+	@Test
+	public void downloadJsonFileFromOneDrive() throws Exception {
+		final InputStream stream = testBase.graphClient.me()
+		.drive()
+		.root()
+		.itemWithPath("test.json")
```
nikithauc, line 95: Where is this test.json file located?

src/main/java/com/microsoft/graph/core/Constants.java
```diff
@@ -17,6 +17,10 @@ private Constants() {
 	 * The content type for JSON responses
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
+	/**
+	 * The content type for text responses
+	 */
+	public static final String TEXT_CONTENT_TYPE = "text/plain";
```
baywet, line 23:
```suggestion
```

src/main/java/com/microsoft/graph/core/Constants.java
```diff
@@ -17,6 +17,10 @@ private Constants() {
 	 * The content type for JSON responses
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
+	/**
+	 * The content type for text responses
+	 */
```
baywet, line 22:
```suggestion
```

src/main/java/com/microsoft/graph/core/Constants.java
```diff
@@ -17,6 +17,10 @@ private Constants() {
 	 * The content type for JSON responses
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
+	/**
+	 * The content type for text responses
```
baywet, line 21:
```suggestion
```

src/main/java/com/microsoft/graph/core/Constants.java
```diff
@@ -17,6 +17,10 @@ private Constants() {
 	 * The content type for JSON responses
 	 */
 	public static final String JSON_CONTENT_TYPE = "application/json";
+	/**
```
baywet, line 20:
```suggestion
```

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
baywet:
```suggestion
							contentType.contains(Constants.JSON_CONTENT_TYPE)) {
```

src/main/java/com/microsoft/graph/http/CoreHttpProvider.java
baywet: I agree on the fact we could/should break that method further down in smaller pieces. However I don't want to introduce too many changes at a time for traceability purposes. Extracting getHttpRequest already divided the method's length by 2 :)

src/test/java/com/microsoft/graph/functional/OneDriveTests.java
baywet: this unit test assumed it'd already be here. I was unclear on how we arrange our integration testing environment so I didn't want to make the "test too big for no reason". But following the discussion we had on Monday, it's clear that we don't have something clearly defined yet, so I'll keep this in mind and write tests that are a little more self sufficient, even if there is redundancy. I just added arragement at the beginning of the test to upload the file we'll need to download.

zengin (approved):

PR #481, opened 2020-09-17
bugfix/chunk upload outlook

### LINKED ISSUE #423: ChunkedUploadProvider for AttachmentItem halts after first chunk
Uploading chunks return reponse code 200 where the ChunkedUploadResponseHandler awaits 202. That finishes the chunked upload immediately and the upload will not finish with a 201 Created.
You can get it to work if you extend the ChunkedUploadResponseHandler and overwrite the Response Code and result return:
package com.example.ms.graph;

```Java
import java.io.BufferedInputStream;
import java.io.IOException;
import java.io.InputStream;

import com.microsoft.graph.concurrency.ChunkedUploadResponseHandler;
import com.microsoft.graph.http.DefaultHttpProvider;
import com.microsoft.graph.http.GraphServiceException;
import com.microsoft.graph.http.HttpResponseCode;
import com.microsoft.graph.http.IConnection;
import com.microsoft.graph.http.IHttpRequest;
import com.microsoft.graph.logger.ILogger;
import com.microsoft.graph.models.extensions.AttachmentItem;
import com.microsoft.graph.models.extensions.UploadSession;
import com.microsoft.graph.requests.extensions.ChunkedUploadResult;
import com.microsoft.graph.serializer.ISerializer;

import okhttp3.Response;

public class ChunkedUploadResponseHandler2<UploadType> extends ChunkedUploadResponseHandler<UploadType> {

	private final Class<UploadType> deserializeTypeClass;

	public ChunkedUploadResponseHandler2(Class<UploadType> deserializeTypeClass) {
		super(deserializeTypeClass);
		this.deserializeTypeClass = deserializeTypeClass;
	}

	/**
	 * Generate the chunked upload response result
	 *
	 * @param request
	 *            the HTTP request
	 * @param response
	 *            the HTTP response
	 * @param serializer
	 *            the serializer
	 * @param logger
	 *            the system logger
	 * @return the chunked upload result, which could be either an uploaded item or error
	 * @throws Exception
	 *             an exception occurs if the request was unable to complete for any reason
	 */
	@Override
	public ChunkedUploadResult<UploadType> generateResult(final IHttpRequest request, final Response response, final ISerializer serializer, final ILogger logger)
			throws Exception {
		InputStream in = null;
		try {
			if (response.code() == HttpResponseCode.HTTP_ACCEPTED || response.code() == HttpResponseCode.HTTP_OK) {
				logger.logDebug("Chunk bytes has been accepted by the server.");
				in = new BufferedInputStream(response.body().byteStream());
				final UploadSession session = serializer.deserializeObject(DefaultHttpProvider.streamToString(in), UploadSession.class);

				return new ChunkedUploadResult<UploadType>(session);

			}
			else if (response.code() == HttpResponseCode.HTTP_CREATED) { // || response.code() == HttpResponseCode.HTTP_OK
				logger.logDebug("Upload session is completed, uploaded item returned.");
				// in = new BufferedInputStream(response.body().byteStream());
				// String rawJson = DefaultHttpProvider.streamToString(in);
				// UploadType uploadedItem = serializer.deserializeObject(rawJson, this.deserializeTypeClass);
				String itemUrl = response.header("Location");
				logger.logDebug(String.format("Upload session is completed, upload url: %s", itemUrl));
				AttachmentItem ai = new AttachmentItem();
				ai.name = itemUrl;
				UploadType uploadedItem = (UploadType) ai;
				return new ChunkedUploadResult<UploadType>(uploadedItem);
			}
			else if (response.code() >= HttpResponseCode.HTTP_CLIENT_ERROR) {
				logger.logDebug("Receiving error during upload, see detail on result error");

				return new ChunkedUploadResult<UploadType>(GraphServiceException.createFromConnection(request, null, serializer, response, logger));
			}
		}
		finally {
			if (in != null) {
				try {
					in.close();
				}
				catch (IOException e) {
					logger.logError(e.getMessage(), e);
				}
			}
		}
		return null;
	}
}
```

This will successfully upload the file and create the resource on the server.
The call will be as stated on the documentation:
```Java
File file = new File(ea.getFilepath());

AttachmentItem attachmentItem = new AttachmentItem();
attachmentItem.attachmentType = AttachmentType.FILE;
attachmentItem.name = file.getName();
attachmentItem.size = file.length();
attachmentItem.contentType = "application/octet-stream";

InputStream fileStream = new FileInputStream(file);
long streamSize = attachmentItem.size;

// Create a callback used by the upload provider
IProgressCallback<AttachmentItem> callback = new IProgressCallback<>() {
    @Override
    // Called after each slice of the file is uploaded
    public void progress(final long current, final long max) {
        logger.logDebug(String.format("Uploaded %d bytes of %d total bytes", current, max));
    }

    @Override
    public void success(final AttachmentItem result) {
        logger.logDebug(String.format("Uploaded file."));
    }

    public void failure(final ClientException ex) {
        logger.logError(String.format("Error uploading file: %s", ex.getMessage()), ex);
    }
};

UploadSession uploadSession = getClient().me().messages(drafted.id).attachments().createUploadSession(attachmentItem).buildRequest().post();
logger.logDebug(String.format("Upload session URL: %s", uploadSession.uploadUrl));

ChunkedUploadProvider<AttachmentItem> chunkedUploadProvider = new ChunkedUploadProvider<>(uploadSession, getClientUnauthorized(), fileStream,
        streamSize, AttachmentItem.class);

// Config parameter is an array of integers
// customConfig[0] indicates the max slice size
// Max slice size must be a multiple of 320 KiB, numRetries
int[] customConfig = {
        320 * 1024, 1
};

// Do the upload
chunkedUploadProvider.upload(callback, customConfig);
```

So it needs to be checked if there is really consistent behaviour on the server side regarding the response codes for accepting chunks when using the different endpoints onedrive || messages || events etc.
Maybe it is a good idea to establish separate reponsehandlers and uploadproviders for the different endpoints, especially when mixing the audience. The preauthenticated uploadurl also seems not to be the ideal solution for a permission handover.
PS: Don't forget to add the permission Mail.ReadWrite in case you upload an attachment to an existing message, otherwise you will end up also in a 401!
Originally posted by @gantners in #409 (comment)
[AB#6035](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6035)

baywet:
Fixes #423

-

### Other links

-


src/main/java/com/microsoft/graph/concurrency/ChunkedUploadResponseHandler.java
```diff
@@ -154,30 +159,35 @@ public void configConnection(final Response response) {
...
-        		return new ChunkedUploadResult<UploadType>(uploadedItem);
-        	} else if (response.code() >= HttpResponseCode.HTTP_CLIENT_ERROR) {
+        	if (response.code() >= HttpResponseCode.HTTP_CLIENT_ERROR) {
         		logger.logDebug("Receiving error during upload, see detail on result error");
 
         		return new ChunkedUploadResult<UploadType>(
         				GraphServiceException.createFromConnection(request, null, serializer,
         						response, logger));
-        	}
+            } else if (response.code() >= HttpResponseCode.HTTP_OK
```
MIchaelMainer, line 168: The `if` accounts for all response codes 400 and higher. This `else if` accounts for all 200s. A final `else` should account for all 300s as being unexpected.

src/main/java/com/microsoft/graph/concurrency/ChunkedUploadResponseHandler.java
```diff
@@ -154,30 +159,35 @@ public void configConnection(final Response response) {
...
         		return new ChunkedUploadResult<UploadType>(
         				GraphServiceException.createFromConnection(request, null, serializer,
         						response, logger));
-        	}
+            } else if (response.code() >= HttpResponseCode.HTTP_OK 
+                    && response.code() < HttpResponseCode.HTTP_MULTIPLE_CHOICES) {
+                final Map<String, String> headers = responseHeadersHelper.getResponseHeadersAsMapStringString(response);
+                final String contentType = headers.get(Constants.CONTENT_TYPE_HEADER_NAME);
+                final String location = headers.get("Location");
+                if(contentType != null
```
MIchaelMainer, line 173: If contenttype and location are null, we should have an `else` where we log this unexpected scenario at the debug level.

src/main/java/com/microsoft/graph/http/HttpResponseHeadersHelper.java
```diff
@@ -0,0 +1,48 @@
...
+	 * @return           the set of headers names and value
+	 */
+	public Map<String, String> getResponseHeadersAsMapStringString(final Response response) {
+		final Map<String, String> headers = new HashMap<>();
+		int index = 0;
+		Headers responseHeaders = response.headers();
+		while (index < responseHeaders.size()) {
+			final String headerName = responseHeaders.name(index);
+			final String headerValue = responseHeaders.value(index);
+			if (headerName == null && headerValue == null) {
```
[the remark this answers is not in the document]
baywet: thanks, I moved the code without looking at it.

src/main/java/com/microsoft/graph/concurrency/ChunkedUploadResponseHandler.java
baywet: there's already a default return at the end of the method. It was returning null but I added a more explicit error message.

src/main/java/com/microsoft/graph/concurrency/ChunkedUploadResponseHandler.java
baywet: added, thanks

MIchaelMainer (approved):

PR #483, opened 2020-09-18
- fixes a bug where collections would not be deserialized properly

### LINKED ISSUE #472: Unable to upload a file
I'm able to get the list of all files in a OneDrive folder and also download a file from a OneDrive Folder to my android device.
But I am unable to upload a file to OneDrive.
It silently crash when I call the post() function.
I use MSAL to connect to OneDrive.
My scopes are: "User.Read", "Files.ReadWrite.All"
I use microsoft-graph:2.0.0

Here is a simplified code with minimal error handling inspired from the sample app found [here](https://docs.microsoft.com/en-us/graph/sdks/large-file-upload?tabs=java):

Note: I can debug step all the way to buildRequest() but it crashes when I try to step thru  .post() 
The requestUrl seams ok: `https://graph.microsoft.com/v1.0/me/drive/root:/TestFolder/Test  File.sco:/microsoft.graph.createUploadSession`
(Crashes at the line that is commented // ************* Silently craches here   ***************** )

The runtime exception cause is: com.microsoft.graph.core.ClientException: Error during http request

```
   private void oneDrive_UploadFile(final IAuthenticationResult authenticationResult, final String strLocalFilePath)
    {
        // Create a callback used by the upload provider
        IProgressCallback<DriveItem> callback = new IProgressCallback<DriveItem>()
        {
            @Override
            // Called after each slice of the file is uploaded
            public void progress(final long current, final long max)
            {
                System.out.println(
                        String.format("Uploaded %d bytes of %d total bytes", current, max)
                );
            }

            @Override
            public void success(final DriveItem result)
            {
                System.out.println(
                        String.format("Uploaded file with ID: %s", result.id)
                );
                executeOneDriveTask(authenticationResult);
            }

            public void failure(final ClientException ex)
            {
                System.out.println(
                        String.format("Error uploading file: %s", ex.getMessage())
                );
            }
        };

        // Get an input stream for the file
        File file = new File(strLocalFilePath);
        if (file.exists())
        {
            final String accessToken = authenticationResult.getAccessToken();

            IGraphServiceClient graphClient =
                    GraphServiceClient
                            .builder()
                            .authenticationProvider(new IAuthenticationProvider()
                            {
                                @Override
                                public void authenticateRequest(IHttpRequest request)
                                {
                                    Log.d(TAG, "Authenticating request," + request.getRequestUrl());
                                    request.addHeader("Authorization", "Bearer " + accessToken);
                                }
                            })
                            .buildClient();

            InputStream fileStream = null;
            try
            {
                fileStream = new FileInputStream(file);
            }
            catch (FileNotFoundException e)
            {
                e.printStackTrace();
            }
            long streamSize = file.length();

            // Create an upload session
            UploadSession uploadSession = graphClient
                    .me()
                    .drive()
                    .root()
                    // itemPath like "/Folder/file.txt"
                    // does not need to be a path to an existing item
                    .itemWithPath(AppPrefs.getInstance().getOneDriveFolder() + '/' + file.getName())
                    .createUploadSession(new DriveItemUploadableProperties())
                    .buildRequest()
                    .post();   // ************* Silently craches here   *****************  

            ChunkedUploadProvider<DriveItem> chunkedUploadProvider = new ChunkedUploadProvider<DriveItem>(
                    uploadSession,
                    graphClient,
                    fileStream,
                    streamSize,
                    DriveItem.class);

            // Config parameter is an array of integers
            // customConfig[0] indicates the max slice size
            // Max slice size must be a multiple of 320 KiB
            int[] customConfig = {2 * 320 * 1024};

            // Do the upload
            try
            {
                chunkedUploadProvider.upload(callback, customConfig);
            }
            catch (IOException e)
            {
                e.printStackTrace();
            }
        }
    }

```

Any clue ?

Thanks

baywet:
Fixes #472

- adds support for deserialization of pagecollections (only attachements were supported before for some reason)

### Other links

-


src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
```diff
@@ -0,0 +1,138 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonParser;
+import com.google.common.reflect.TypeToken;
+import com.microsoft.graph.http.BaseCollectionPage;
+import com.microsoft.graph.http.IRequestBuilder;
+import com.microsoft.graph.logger.ILogger;
+
+// those imports are useless but build will fail if code-gen conventions change, keep it
```
MIchaelMainer, line 41: Is this comment still applicable? If not, please delete these imports.

src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
```diff
@@ -0,0 +1,138 @@
...
+	@SuppressWarnings("unchecked")
+	public static <T1, T2 extends IRequestBuilder> BaseCollectionPage<T1, T2> deserialize(final JsonElement json, Type typeOfT, final ILogger logger) throws JsonParseException {
+		if (json == null) {
+			return null;
+		}
+		serializer = new DefaultSerializer(logger);
+		final JsonObject[] sourceArray = serializer.deserializeObject(json.toString(), JsonObject[].class);
+		final ArrayList<T1> list = new ArrayList<T1>(sourceArray.length);
+		final String collectionPageClassCanonicalName = typeOfT.getTypeName();
+		final String entityClassCanonicalName = collectionPageClassCanonicalName
```
MIchaelMainer, line 108: I don't understand this string manipulation. Can you add a comment to explain why this is done? This occurs in a few places so this comment applies to those places as well.

src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
```diff
@@ -0,0 +1,138 @@
...
+					.substring(0, collectionPageClassCanonicalName.length() - pageLength - collectionLength)
+					.replace("requests", "models");
+		try {
+			final Class<?> entityClass = Class.forName(entityClassCanonicalName);
+			for (JsonObject sourceObject : sourceArray) {
```
MIchaelMainer, line 113: If this is a heterogenous collection where the collection is defined as collection of the basetype, the call to deserialize will  deserialize to the basetype, putting all non-base type properties into additional data property bag, correct? 

If I understand this correctly, we may want to query JsonObject for odata.type, and use that to extract and specify the entityClass

src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
baywet: updated the comment to make it a bit more self-explanatory

src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
baywet: added comments

src/main/java/com/microsoft/graph/serializer/CollectionPageSerializer.java
baywet: thanks! I hadn't thought about this case. I implemented support for heterogenous collections and also fixed a bug where the default deserializer would not support namespaces.

MIchaelMainer (approved):

PR #503, opened 2020-09-25
- updaates generated files for orderby support

### LINKED ISSUE #502: missing support for orderBy odata query parameter
today using orderBy requires to new a query option, which is verbose. We can facilitate this by adding a fluent method like for filter and select

baywet:
fixes #502


zengin (approved):

giulianf:
Hi,
This feature is not merged to "master" branch ?
Thx

PR #514, opened 2020-10-01
feature/pom-placeholder

baywet:
This adds a package file as a place holder to enable the dependency graph https://github.com/microsoftgraph/msgraph-sdk-java/network/dependencies
		as the dependency graph is not compatible with gradle https://docs.github.com/en/free-pro-team@latest/github/visualizing-repository-data-with-graphs/about-the-dependency-graph#supported-package-ecosystems

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,4 @@ hs_err_pid*
 
 # Maven
 /target/
-/pom.xml
 local.properties
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,14 @@ java {
     modularity.inferModulePath = true
 }
 
+sourceSets {
+    main {
+        java {
+            exclude 'pom.xml'
+        }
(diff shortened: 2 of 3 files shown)


zengin (changes requested):
please provide a description

baywet:
@zengin yep, sorry, this one was created before our last discussion

zengin (approved):

MIchaelMainer:
Does this mean that you have to manually update the pom.xml whenever you make a change to the dependencies or is there a tool/script to do this?

baywet:
manual at this point, but again this is only used for the dependency tree of GitHub. This way GitHub is able to tell "oh this repo actually publishes a lib, let's see what it depends on and what depends on it to build the tree". Worst case if we forget to update it, our dependency tree will be inaccurate.

PR #520, opened 2020-10-05
- fixes a regression where the content method would be missing for message

### LINKED ISSUE #323: Getting the mime content of a message is not implement in the java sdk
According to https://docs.microsoft.com/de-de/graph/outlook-get-mime-message it's possible to get the mime content representation of a mail.
This works fine using the graph explorer with this url:
/me/messages/{id}/$value
The implementation is missing in the java api.
Or is there an other way to get the mime content? I didn't find anything.
Thanks, Christoph
[AB#6051](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6051)

baywet:
fixes #323

### DIFF
--- a/src/main/java/com/microsoft/graph/requests/extensions/IMessageRequestBuilder.java
+++ b/src/main/java/com/microsoft/graph/requests/extensions/IMessageRequestBuilder.java
@@ -58,6 +58,8 @@ public interface IMessageRequestBuilder extends IRequestBuilder {
     ISingleValueLegacyExtendedPropertyCollectionRequestBuilder singleValueExtendedProperties();
 
     ISingleValueLegacyExtendedPropertyRequestBuilder singleValueExtendedProperties(final String id);
+
+    IMessageStreamRequestBuilder content();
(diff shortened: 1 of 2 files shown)


zengin (approved):
@baywet are these handcrafted files? If generated, do we have a corresponding generator PR?

Can you share the original PR that introduced the regression? That can help us catch similar regressions in subsequent PRs. Thanks!

baywet:
@zengin the PR that introduced the regression was the one with the order by.
The reason it introduced the regression is because I was working off a metadata version that didn't have the hasstream on the message entity when working on the order by one. And then the merge sequence of the different PRs kind of derailed it (I've probably started the order by PR from a branch that had the code for the message mime content).
No changes required in either the generator or the metadata, all the changes are here, it's just bad sequencing on my end...

baywet:
and I generated those changes, it's not a manual change.

zengin:
Thanks for the explanation!

PR #526, opened 2020-10-05
feature/no interfaces

baywet:
removes generated interfaces (package size reduction, simplicity, build time...)


MIchaelMainer (approved):
Glorious removal. Sampled review.

baywet:
Just with that the lib binary size went from 7.6MB to 5.9

PR #558, opened 2020-11-12
feature/android api baseline

### LINKED ISSUE #542: Getting crash in Android OS 7
While in process of uploading file to OneDrive, we are getting crash on device OS 7. **We couldn't verify this crash on Android OS level below 7 as we get the crash in CalenderSerializer before this flow kicks in.**

Crash logs are as follow:
--------- beginning of crash 10-15 10:32:12.228 13689 13689 E AndroidRuntime: FATAL EXCEPTION: main 10-15 10:32:12.228 13689 13689 E AndroidRuntime: Process: com.adobe.reader, PID: 13689 10-15 10:32:12.228 13689 13689 E AndroidRuntime: java.lang.NoSuchMethodError: No interface method getTypeName()Ljava/lang/String; in class Ljava/lang/reflect/Type; or its super classes (declaration of 'java.lang.reflect.Type' appears in /system/framework/core-oj.jar) 10-15 10:32:12.228 13689 13689 E AndroidRuntime: at com.microsoft.graph.serializer.CollectionPageSerializer.deserialize(CollectionPageSerializer.java:115) 10-15 10:32:12.228 13689 13689 E AndroidRuntime: at com.microsoft.graph.serializer.GsonFactory$12.deserialize(GsonFactory.java:229) 10-15 10:32:12.228 13689 13689 E AndroidRuntime: at com.microsoft.graph.serializer.GsonFactory$12.deserialize(GsonFactory.java:224) 10-15 10:32:12.228 13689 13689 E AndroidRuntime: at 

### Expected behavior
It shouldn't crash and handle newer apis with fallback mechanish or apis that is supported in older OS version

### Devices tested where issue is observed:
Asus Tab, Samsung J7, Samsung Tab A, Galaxy S6

### Cause
After some analysis, I found that crash is originating from CollectionPageSerializer which is being registered into Gson type adapter and which in function deserialize calls the method Type#getTypeName() which is a default method added after Java 1.8

Also can we make gson variable in DefaultSerializer public so that while configuring the graph client we can replace these problematic TypeAdapters with the custom solution. Like in case of when XXX is not supported in simpleDateFormat then same can be replaced with joda-time as we already have that dependency. Similarly in this case as well usage of getTypeName() can be replace with getName() or getCannonicalName() as to avoid crash

### LINKED ISSUE #546: Clearly define and document minimum support API level
related #542
related #286 
related #530

There's a number of instances where we are using runtime APIs that are not available in Android, or only available for specific API levels and up. Saying "this library requires java X" is not enough as not all java 8 APIs are available on android to this day and desugaring won't bring those APIs to a specific API level.
During this evaluation we should consider downstream libraries, market shares of different android versions, and code of "writing legacy code/workarounds".
Once an API level is selected, we need to hunt for any transgression and set tooling to check for use of APIs unavailable for a specific API level. (static analysis)
This should be possible via the android configuration in build.gradle which should warn about any [transgression once configured](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd).
 
[AB#6385](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6385)

baywet:
fixes #546
fixes #542
- upgrades gradle wrapper
- fixes a bug where locale wasn't definied for string formats potentially leading to formatting errors
- adds gradle wrapper for android profile
- adds gitignore for android test project
- adds android project structure
- adds nonnull and nullable annotations
- fixes use of type.getTypeName API not available in level 21 through 25
- adds failsafe for X character not available in level 21 through 23
- removes synthetic accessors

Note:

[template used for library](https://github.com/jaredsburrows/android-gradle-java-library-template)

TODO:
- [x] mutualize the dependencies through a common gradle file + apply from: path to file
- [x] add github actions pipelines to run this on PRs
- [x] fix remaining issues
- [x] decide which API level to adopt (probably 21 or 23)
- [x] update documentation
- [x] reflect changes on core


zengin (approved):
I did a sampled review, LGTM.

PR #560, opened 2020-11-18
- adds type summary for easier reviews

baywet:
this pull request adds a type summary application to make reviews easier, similar to what we have in dotnet today.
It also adds a github actions workflows to trigger that summary generation on pushes to java files.


typesummary/app/src/main/java/typesummary/App.java
```diff
@@ -0,0 +1,125 @@
...
+            if(!methodsNameToSkip.contains(method.getName())) {
+                writer.write("method " + method.getName(), 2);
+                writer.write("return type " +method.getReturnType().getName(), 4);
+                serializeParameters(method, writer);
+            }
+        }
+    }
+    private static void serializeParameters(final Method method, final ILogWriter writer) {
+        for(Parameter parameter: method.getParameters()) {
+            writer.write("param " + parameter.getName() + " : " + parameter.getType().getName(), 4);
```
zengin, line 122: Is it expected that `parameter.getName()` only returns `arg0`, `arg1` etc.?

zengin (approved):

typesummary/app/src/main/java/typesummary/App.java
baywet: do you mean it's never returning the actual parameter name? I'll take a look, I think I missed that. thanks!

zengin: yeah, typeSummary.txt has only those as parameter names.

baywet: so this is actually a side effect caused by the fact that the SDK is technically still building targeting java 7 at this point (v3 upgrades to java 8). I changed the project structure so the typesummary tool builds the SDK itself (instead of 2 separate steps) under java 8 and now the parameters names show up.

The CI should update the summary text with the parameters names now.
some more reading https://github.com/paul-hammant/paranamer

baywet:
@zengin a follow up question I had was : are we interested by those methods (that are on every object) `additionalDataManager`, `getRawObject`, `setRawObject` ?

zengin:
@baywet if we are not changing where they are implemented (directly or in the parent) and if we don't anticipate them changing in the future, I am OK with not inflating the summary file with repeated methods.

baywet:
hum, I think there are a few places in the generic refactoring where I moved those up, let's keep it then, thanks!

baywet:
do you mind approving now?

MIchaelMainer (approved):

zengin (approved):

PR #564, opened 2020-11-19
- fixes a bug where the type summary would not be deterministic

baywet:


typesummary/app/src/main/java/typesummary/App.java
```diff
@@ -106,9 +106,26 @@ private static void serializeFields(final Class<?> clazz, final ILogWriter write
         }
     }
     private static List<String> methodsNameToSkip;
+    private static String delimiter = " ";
     private static void serializeMethods(final Class<?> clazz, final ILogWriter writer) {
         final Method[] methods = clazz.getMethods();
-        Arrays.sort(methods, (o1, o2) -> o1.getName().compareTo(o2.getName()));
+        Arrays.sort(methods, (o1, o2) -> (o1.getName() +
```
zengin, line 112: nit: can you please add a short comment on why we do this :)

zengin (approved):

typesummary/app/src/main/java/typesummary/App.java
baywet: done!

zengin (approved):

PR #569, opened 2020-11-25
- fixes #568: a bug where time duration deserialization would fail because of quotes

### LINKED ISSUE #568: Getting NULL on deserializing the metrics in Call Record of type "Duration"
### Expected behavior
Should get the values for the metrics like AverageRoundTripDelay, AverageJitter, and Max Jitter which are of type "Duration".

### Actual behavior
Metrics like AverageRoundTripDelay, AverageJitter, MaxJitter which are of type "Duration" are always deserialized as NULL instead of the value reported in the call record. In GsonFactory.java at line:208, it should be "getAsString()" and not "toString()". 
This is preventing from getting the values.

baywet:
fixes #568 a bug where duration deserlization would fail because of quotes

### DIFF
--- a/src/main/java/com/microsoft/graph/serializer/GsonFactory.java
+++ b/src/main/java/com/microsoft/graph/serializer/GsonFactory.java
@@ -205,7 +205,7 @@ final class GsonFactory {
                                        final Type typeOfT,
                                        final JsonDeserializationContext context) throws JsonParseException {
                 try {
-                    return DatatypeFactory.newInstance().newDuration(json.toString());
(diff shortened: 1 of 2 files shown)


darrelmiller:
Does this fix need to happen in core too?  Changing service library code by hand makes me queasy.

baywet:
@darrelmiller the game plan it to move everything that can be moved from service lib to core lib once I'm done with all the fixes and code gen changes.

baywet:
and before I do that I'll merge dev into feature/v3 (which I do on regular basis to make sure we don't have regressions)

darrelmiller (approved):

PR #581, opened 2020-12-01
feature/editor-config

baywet:
I've been fighting with tabs and spaces for a while now. This comes from the very good suggestion from @pschaeflein

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[*.java]
+indent_style = tab
+insert_final_newline = true
+tab_width = 3
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 [*.java]
-indent_style = tab
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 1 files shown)


zengin:
This is useful for handcrafting files in this repo, but I don't think it will fix the auto-generated files.

I think we should start from MSGraph-SDK-Code-Generator repo. Most code we generate there is using spaces (width: 4). I would suggest setting the standard to that for minimal effect on file history.

zengin (approved):

baywet:
@zengin thanks for the suggestion. I've added a configuration for tt files as well on the generator. https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/336
It only seems to work when opening [VS in folder mode](https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes-v15.0#coding-convention-support-through-editorconfig) and for the part that's not specific (only the CSharp that's not part of the template text gets indented). But it should come handy as we switch to dotnet-core and vs code for the generator.

PR #584, opened 2020-12-03
feature/count

baywet:
- codegen update: adds count odata query option
- adds base methods for odata query option count
- removes extra code gen that shouldn't have been generated
- removes the unused app root constant - adds a constant for text rutun type
- adds a method to read the odata count
- adds the ability to parse raw text responses
- adds classes for primitive response types
- code-gen update: raw count support
- fail safe in case we can't parse the long
- adds test for raw and query count


src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+import com.microsoft.graph.concurrency.ICallback;
+import java.util.Arrays;
+import java.util.EnumSet;
+import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
+import com.microsoft.graph.core.IBaseClient;
+import com.microsoft.graph.http.BaseRequestBuilder;
+
+/**
+ * The class for the Call Record Request Builder.
```
zengin, line 36: Comment doesn't match.

src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+    public PrimitiveRequestBuilder(@Nonnull final String requestUrl, @Nonnull final IBaseClient client, @Nullable final java.util.List<? extends com.microsoft.graph.options.Option> requestOptions, @Nonnull final Class<T> primitiveTypeClass) {
+        super(requestUrl, client, requestOptions);
+        this._returnTypeClass = primitiveTypeClass;
+    }
+
+    /**
+     * Creates the request
+     *
+     * @param requestOptions the options for this request
+     * @return the CallRecordRequest instance
```
zengin, line 58: Comment doesn't match.

src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
```diff
@@ -0,0 +1,75 @@
+// ------------------------------------------------------------------------------
+// Copyright (c) 2017 Microsoft Corporation
```
zengin, line 2:
```suggestion
// Copyright (c) 2020 Microsoft Corporation
```

src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
baywet: good catch, thanks. Fixed.

src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
baywet: good catch, thanks. Fixed.

src/main/java/com/microsoft/graph/http/PrimitiveRequestBuilder.java
baywet: fixed manually with other comments issues

src/main/java/com/microsoft/graph/requests/extensions/WorkbookChartSeriesCollectionRequestBuilder.java
```diff
@@ -21,7 +21,7 @@
 import com.microsoft.graph.requests.extensions.WorkbookChartSeriesItemAtRequestBuilder;
 import com.microsoft.graph.http.BaseCollectionRequestBuilder;
 import com.microsoft.graph.core.IBaseClient;
-
+import com.microsoft.graph.http.PrimitiveRequestBuilder;
```
MIchaelMainer, line 24: This probably shouldn't have this import as it isn't used and I don't think it makes sense here.

MIchaelMainer (approved):
I like the default count().

src/main/java/com/microsoft/graph/requests/extensions/WorkbookChartSeriesCollectionRequestBuilder.java
MIchaelMainer: I don't know why it is here, although there is a count method composed

baywet: fixed. I wanted to keep the template simpler but inserting references when not needed can be confusing. Plus it has a (limited) impact on build performance for us

zengin (approved):

PR #607, opened 2021-01-05
Generated from new pipeline using the existing clean metadata

MIchaelMainer:
Let's take this in before we do a full run with capturing and cleaning metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ repositories {
 
 dependencies {
     // Include the sdk as a dependency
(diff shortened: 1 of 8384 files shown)


baywet (dismissed):

baywet:
likewise, the following command brings no other changes than the version ones after reverting the security casing thing.
```Shell
git diff dev -I "\/\/\sTemplate Source:\s\w+\.java\.tt" --ignore-space-change
```

baywet (approved):

PR #610, opened 2021-01-13
feature/future

### LINKED ISSUE #533: evaluate the possibility of Future API surface
Context: java's equivalent to Promises/Tasks are Futures. Adding an API surface that returns Futures (and potentially removing the callback one??) would enable our customers to write modern Java in their applications

I suppose so but my callback wrapper is pretty straightforward; it wasn't a huge deal writing it.

```scala
object ICallbackWrapper extends StrictLogging {

  private def cbF[T](): (Future[T], ICallback[T]) = {
    val p = Promise[T]()
    val cb = new ICallback[T] {
      override def success(result: T): Unit = {
        val _ = p.trySuccess(result)
      }

      override def failure(ex: ClientException): Unit = {
        val _ = p.tryFailure(ex)
      }
    }

    (p.future, cb)
  }

  def toFut[T](fun: ICallback[T] => Unit)(
      implicit
      ec: ExecutionContext,
  ): Future[T] = {
    val (f, cb) = cbF[T]()

    Limiter.apiFut {
      () =>
        fun(cb)
        f
    }
  }

  def toFut[V, T](v: V, fun: (V, ICallback[T]) => Unit)(
      implicit
      ec: ExecutionContext,
  ): Future[T] = {
    val (f, cb) = cbF[T]()

    Limiter.apiFut {
      () =>
        fun(v, cb)
        f
    }
  }

  def uploadFut[V, T](v: V, len: Long, fun: (V, ICallback[T]) => Unit)(
      implicit
      ec: ExecutionContext,
  ): Future[T] = {
    val (f, cb) = cbF[T]()

    Limiter.attachFut(
      () => {
        fun(v, cb)
        f
      },
      len,
    )
  }
}
```

By far the biggest issue with using graph are the very restrictive rate limits... I shimmed a "rate limiter" into the above wrapper so I'm able to limit both number of simultaneous connections as well as ensure uploads don't go over the 15mb/30 sec limit.  It took a lot of trial and error getting the futures passed just right so I didn't go over the max 4 connection cap, but it seems to be working well in practice.

_Originally posted by @lespea in https://github.com/microsoftgraph/msgraph-sdk-java/issue_comments/705202102_ 
[AB#6315](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6315)

baywet:
- codegen update: future API surface
- replaces callback api surface by future API surface
- bumps minumum API level version number to 26

fixes #533
replaces the callback API surface by futures for async operations

TODO:
- [x] update readme for new API level requirements

Here is an example of the experience it enables:
Before
```Java
final ICallback<Subscription> callback = new ICallback<Subscription> {
   @Override
   public void succes(final Subscription result) {
        System.Out.println("created " + s.id);
   }
   @Override
   public  void failure(final ClientException ex) {
   //log errror
  }
};

client.subscriptions()
    .buildRequest()
    .post(mySub, callback);
```
After 
```Java
client.subscriptions()
    .buildRequest()
    .futurePost(mySub)
    .thenApply(s ->System.Out.println("created " + s.id))
    .get();
```


nikithauc:
For my learning, why rename the HTTP request methods such as send() or post() to futureSend() or futurePost().
Is the 'future' prefix an indicator of the return type or a good practice?
Would using the HTTP request methods without the prefix be more conventional?

baywet:
@nikithauc thanks for reviewing the PR.
The reason why I had to set a different name is because now that those methods don't have a parameter for the callback, they conflict with the synchronous methods. I did some research to try to identify whether there was a convention like in dotnet with the "Async" suffix, but it doesn't seem to be the case. So I went with the prefix. Changing the prefix/suffix should only be a matter of a big string replace in the templates and in the "manually updated files".
Please let me know if you hear about some kind of convention for Java.

baywet:
as a follow up, I decided to ask the question to the java community directly, let's see what comes back https://stackoverflow.com/questions/65737878/java-naming-conventions-for-methods-returning-futures

MIchaelMainer:
Since we're returning a CompletableFuture, I don't think we need "future" in the method name. I do like having Async in the method name like commenter in the SO post said. But then I'm biased. We should keep the sync functions and mark them as obsolete, while gaining the futures based methods.

baywet:
Yeah, I'll wait a couple of more days before merging to see if anything else pops as an answer to my question on SO.
As for the sync API, I already argued with @darrelmiller to simply remove it as we're shipping a new major release (which would make the lib much smaller, and solve the naming issue) but he said no.

MIchaelMainer:
For reviewer convenience:
`sdvdiff 30586eafb3ee1d77b5811285521f18a5b0e45f1c 6a5765eeecdd992e3e7dd4f8a0a252edb9440cbf`

src/main/java/com/microsoft/graph/concurrency/IProgressCallback.java
```diff
@@ -24,14 +24,12 @@
 
 /**
  * A callback that describes how to deal with success, failure, and progress
- * 
- * @param <Result> the result type of the successful action
  */
-public interface IProgressCallback<Result> extends ICallback<Result> {
+public interface IProgressCallback {
```
MIchaelMainer, line 28: do we still need this since we remove dthe callback?

src/main/java/com/microsoft/graph/core/BaseClient.java
```diff
@@ -1,16 +1,16 @@
 // ------------------------------------------------------------------------------
 // Copyright (c) 2017 Microsoft Corporation
-// 
+//
```
MIchaelMainer, line 3: Maybe I forgot something but do we still need BaseClient, GraphServiceClioent and IBaseClient here? Aren't we supposed to use core with this?

MIchaelMainer: Same question for the files in the http namespace.

MIchaelMainer (commented):
Sampled through the other generated changes. Looks good.

src/main/java/com/microsoft/graph/concurrency/IProgressCallback.java
baywet: We still need it at one place for upload session to report progress. This way of consumers want to update ui they can

src/main/java/com/microsoft/graph/core/BaseClient.java
baywet: This pr doesn't move anything, it's #615 that moves things into core. Until 615 we did have a lot of duplication and extra complexity.

PR #615, opened 2021-01-19
feature/moving nongen

### LINKED ISSUE #578: move remaining non-generated classes to core
after some initial cleanup (eg #516, #516), we still have classes that are not being generated living in the service library that should be moved to core to avoid duplicated with the beta service library.
This issue tracks that missing work 
[AB#7127](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7127)

baywet:
- moves all non-gen code to core and most of unit tests
- implements IUploadSession for session
- moves the constants out of the core to avoid module naming conflict - renames the graphserviceclient - removes the graph service client interface to avoid shipping breaking changes
- adds a lean builder for the graph service client
- fixes wrong imports
- updates version bump script to match new location
- renames testing package to avoid conflicts
- codegen: removes extension and generated paths
- replaces imports in tests and tools following generated and extensions removal
- codegen update: missing imports and missing comments
- adds a version provider in the client

fixes #578

> Note: CI will fail until the core library's snapshot is published


baywet:
FYI here is [the diff](https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/37206..snippet-generation/37633?expand=1) (with datetime) and the [raptor build](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=37634&view=ms.vss-test-web.build-test-results-tab)

zengin (approved):
I didn't individually removed the files here. Approving this based on Raptor results, snippets diff and the Generator repo test cases.

PR #617, opened 2021-01-21
Don't use reference equality on Integers

tbelaire:
This is a comparison using reference equality instead of value equality. Reference equality of boxed primitive types is usually not useful, as they are value objects, and it is bug-prone, as instances are cached for some values but not others.


baywet (changes requested):
@tbelaire thanks for the contribution!
instead of changing the equality check, would you mind changing the variables types to their primitive equivalents please?
the size() methods return primitive types anyway. This we'll skip the casting into an object and save some tiny memory space.

baywet (approved):

baywet:
thanks for making the changes!

PR #630, opened 2021-01-28
Update generated files with build 38600

MIchaelMainer:


typeSummary.txt
```diff
@@ -69479,9 +69493,9 @@
      param httpMethod : com.microsoft.graph.http.HttpMethod
  class com.microsoft.graph.requests.extensions.GroupRequestBuilder : com.microsoft.graph.http.BaseRequestBuilder
    method acceptedSenders
-     return type com.microsoft.graph.requests.extensions.IDirectoryObjectCollectionRequestBuilder
+     return type com.microsoft.graph.requests.extensions.IDirectoryObjectCollectionWithReferencesRequestBuilder
```
baywet, line 69496: @zengin This could be considered a breaking change but I guess it wasn't working before this change and that's why you added a fix for metadata? https://github.com/microsoftgraph/msgraph-metadata/pull/58

baywet (changes requested):
blocking until we get more details on the accept senders change

typeSummary.txt
baywet: discussed on the dotnet equivalent PR, this is not breaking as it was broken to begin with (nobody could've taken a dependency on it)

baywet (approved):

PR #671, opened 2021-02-19
If an error is available during upload report this in the ClientExcep…

AllenConquest:
During testing of the ChunkedUploadProvider.upload() method, an error in the callback was being created. However, this error is being swallowed by the ChunkedUploadRequest class, and not reported back through the exception handling. This change is to pass the original error back to the calling code. This would be much more helpful in debugging errors occurring in the callback handling.

### DIFF
--- a/src/main/java/com/microsoft/graph/requests/extensions/ChunkedUploadRequest.java
+++ b/src/main/java/com/microsoft/graph/requests/extensions/ChunkedUploadRequest.java
@@ -111,8 +111,8 @@ public class ChunkedUploadRequest {
 
         if (result != null && result.chunkCompleted()) {
             return result;
-        } else 
+        } else
             return new ChunkedUploadResult<UploadType>(
-                new ClientException("Upload session failed.", null));
+                new ClientException("Upload session failed.", result != null ? result.getError() : null));
     }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/graph/requests/extensions/ChunkedUploadRequest.java
```diff
@@ -111,8 +111,8 @@ public ChunkedUploadRequest(final String requestUrl,
 
         if (result != null && result.chunkCompleted()) {
             return result;
-        } else 
+        } else
             return new ChunkedUploadResult<UploadType>(
-                new ClientException("Upload session failed.", null));
+                new ClientException("Upload session failed.", result != null ? result.getError() : null));
```
baywet, line 116:
```suggestion
                new ClientException("Upload session failed.", result == null ? null : result.getError()));
```
nit: readability

baywet (changes requested):
Thank you so much for your feedback and your contribution!
A minor comment here.

baywet (approved):
additional review after implementing the minor change.

andrueastman (approved):
👍

baywet:
@AllenConquest your contribution will go out with version 2.9.0 which should go out next week. Thanks!

AllenConquest:
Thanks for the speedy update. However, I now have an issue trying to upgrade from 2.3.1 to 2.8.1. I will raise a separate issue.

AllenConquest:
Do you have any idea when version 2.9.0 will be release?

baywet:
the plan was to have it released this week. But the service didn't release any update so we decided not to release to save people from an upgrade. I'm tracking the target date [here](https://github.com/microsoftgraph/msgraph-sdk-java/milestone/15)

PR #850, opened 2021-08-17
Bug Fixes

ramsessanchez:
Removed the reference to templates/download-artifacts.yml as it no longer exists in our repo. 
File path management for the Gradle tasks running in the Preview and Release stages
Added Install-Java task to the other stages as well.


.azure-pipelines/buildAndPackage.yml
```diff
@@ -50,9 +50,8 @@ stages:
      runOnce:
       deploy:
         steps:
-        - template: templates/download-artifacts.yml
         - template: templates/secure-files.yml
```
baywet, line 53: there's something I don't understand for both maven releases stages: if the pipeline doesn't download artifacts and doesn't get the sources + java, how is it going to publish anything?

ramsessanchez: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/deployment-jobs?view=azure-devops#runonce-deployment-strategy
I believe that deployment jobs in .yml automatically download and consume the artifacts. Could you double check to make sure I'm interpreting this correctly?

baywet: you're right, it's actually a path management issue, step is looking at D:\a\1\s\_msgraph-sdk-java-core build and package\drop\gradlew when it should look D:\a\1\Drop\gradlew

baywet (approved):

PR #892, opened 2021-09-20
Queue Workflows on pr creation

ramsessanchez:
Queue the workflows when dependabot creates a pr.

### DIFF
--- a/.github/workflows/create-v1.0-pull-request.yml
+++ b/.github/workflows/create-v1.0-pull-request.yml
@@ -50,6 +50,7 @@ jobs:
     needs: create-pull-request
     runs-on: ubuntu-latest
     steps: 
+      - uses: actions/checkout@v2.3.4
       - name: Queue Api Level Lint
         uses: benc-uk/workflow-dispatch@v1
         with:
--- a/.github/workflows/create-v1.0-pull-request.yml
+++ b/.github/workflows/create-v1.0-pull-request.yml
@@ -61,13 +61,13 @@ jobs:
         uses: benc-uk/workflow-dispatch@v1
         with:
           workflow: "CodeQL"
-          toke: ${{ secrets.PERSONAL_TOKEN }}
+          token: ${{ secrets.PERSONAL_TOKEN }}
           ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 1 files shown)


zengin (commented):
Curious to know: what is the reason for these not being automatically queued? I don't see any filters in workflow description: https://github.com/microsoftgraph/msgraph-sdk-java/blob/dev/.github/workflows/codeql-analysis.yml#L9

.github/workflows/create-v1.0-pull-request.yml
```diff
@@ -45,6 +45,30 @@ jobs:
...
         bin/hub pull-request -b "$BASE" -h "$GITHUB_REF" -m "$MESSAGE_TITLE" -m "$MESSAGE_BODY" -r "$REVIEWERS" -a "$ASSIGNEDTO" -l "$LABELS"
+    
+  queue-workflows:
+    needs: create-pull-request
+    runs-on: ubuntu-latest
+    steps: 
+      - name: Queue Api Level Lint
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: "Checks the SDK only using APIs from the targeted API level"
```
baywet, line 56: > Required. The name or ID of the workflow to trigger and run. This is the name declared in the YAML, not the filename
[reference](https://github.com/benc-uk/workflow-dispatch#workflow)

baywet (changes requested):
@zengin  the generation PR is created by a BOT (the default token from github) which doesn't trigger any other workflow (the default token can't trigger workflows, they did that to protect the platform and avoid cascading/infinite loops)

.github/workflows/create-v1.0-pull-request.yml
ramsessanchez: That is the current name on the workflow

ramsessanchez: Should I include a name change to this workflow in this PR,  for the sake of being more concise?

baywet: sorry, I read that sentence the other way around...

baywet (approved):

PR #936, opened 2021-11-16

ramsessanchez:

 The commit hash is _13c430c727d1b6c94deafddeaa7acd24bdf1712b_. 


baywet (changes requested):
[5-7-0_to_java.zip](https://github.com/microsoftgraph/msgraph-sdk-java/files/7546608/5-7-0_to_java.zip)
@ramsessanchez @MIchaelMainer @andrueastman there are two major breaking changes in this week's release:
- all the ediscovery stuff has disappeared, looks like a mistake at the service level to me
- all the reports switches to a binary/stream type, it's possible it's a correction and it didn't use to work before, so maybe we can overlook this one.

baywet:
here's a recap of the discussions I've had with @andrueastman and @darrelmiller on the side since my original comment:

- reports APIs: it's fixing something that was broken and shouldn't be considered a breaking change
- ediscovery APIs: this is [an intentional change](https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads/pullrequest/5300056?path=/Schemas/v1.0-Prod.csdl) and hiding + removing the types from v1 was intended. I'm arguing for a major bump for this change since it removes ~200 types from the SDKs. I'll update here when I get a confirmation.

baywet (approved):
Update : the ediscovery types were accidentally added to v1 too early and only a few weeks ago. Removing them is not considered a breaking change since the APIs behind are not available. We can proceed without a major bump.

baywet (approved):

PR #942, opened 2021-11-18
Update generated files with build 59452 [skip ci]

ramsessanchez:
Highlighting differences based on possible generation suffix fix for Collection entities


baywet (changes requested):
this would translate into a breaking change for the consumers.
also, this would need to be updated to account for the new convention https://github.com/microsoftgraph/msgraph-sdk-java-core/blob/dev/src/main/java/com/microsoft/graph/serializer/DerivedClassIdentifier.java

ramsessanchez:
Closing since changes have been made to disambiguation generation branch.

Issue #258, opened 2019-06-04
Proguard issue with Android release


fiasko131:
Is this sdk really usable on Adroid?

I am referring to this post issue [#213](https://github.com/microsoftgraph/msgraph-sdk-java/issues/213)

[I also posted on Stack Overflow with more details .](https://stackoverflow.com/questions/56445698/issue-with-microsoft-graph-java-sdk-and-proguard-for-android-release)

Please keep us informed as we have been spending time to migrate from [onedrive-sdk-android](https://github.com/OneDrive/onedrive-sdk-android) which seems no longer maintained and no longer works with Pie.

Thanks

**Steps to reproduce the behavior**
Import[ android-java-connect-sample](https://github.com/microsoftgraph/android-java-connect-sample) 
Replace the app gradle file by this:
````java
apply plugin: 'com.android.application'

android {
    compileSdkVersion 28
    flavorDimensions "default"
    defaultConfig {
        applicationId "com.microsoft.graph.connect"
        minSdkVersion 16
        targetSdkVersion 28
        versionCode 2
        versionName "2.0"
        multiDexEnabled true
        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
    }
    signingConfigs {
        release {
            keyAlias 'key0'
            keyPassword 'your password'
            storeFile file('the path of your keyt.jks')
            storePassword 'your password'
        }
        debug {
            keyAlias 'key0'
            keyPassword 'your password'
            storeFile file('the path of your keyt.jks')
            storePassword 'your password'
        }
    }
    buildTypes {
        release {
            minifyEnabled true
            shrinkResources true
            debuggable true
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
            signingConfig signingConfigs.release
            proguardFile 'the path of your proguard-rules.pro'
        }
        debug {
            minifyEnabled false
            shrinkResources false
            signingConfig signingConfigs.debug
            debuggable true
             proguardFile 'the path of your proguard-rules.pro'
        }
    }
    productFlavors {
        legacyDebug {
            minSdkVersion 17
        }
        modernDebug {
            minSdkVersion 21
        }
    }
    testOptions {
        unitTests.returnDefaultValues = true
    }
    configurations {
        all {
            resolutionStrategy.force 'com.android.support:support-annotations:25.3.1'
        }
    }
    dexOptions {
        preDexLibraries = false
        jumboMode = true
[...]

PR #1016, opened 2022-02-24
Update proguard-rules.txt

### LINKED ISSUE #1014: Proguard issue with Android Release
(this is a follow up of #258)

This line appears to include too much code and bloats the size of the build:

```
-keep class com.microsoft.** { *; }
```

I used this line instead:

```
-keepnames public class * implements com.microsoft.graph.serializer.IJsonBackedObject
```

Should the artifact include a consumer-rules.pro so that the config can automatically be exported in release builds?

Digipom:
```-keep class com.microsoft.** { *; }``` keeps too much. I added a new rule that keeps the names of the JSON-backed classes.

Fixes #1014 

- The current Proguard keep rule ```-keep class com.microsoft.** { *; }``` is keeping too much. By keeping just the names of JSON-serializable objects, I was able to reduce the size of the resulting APK by about 600 kb.

### Other links
#258 

- Other notes: Ideally these rules would be included in a "consumer-rules.pro" that gets automatically pulled in by the Android build process. I didn't propose one here because it would affect existing users, so those rules would have to be solidly vetted and ideally part of an automated test suite before being put into place.

### DIFF
--- a/docs/proguard-rules.txt
+++ b/docs/proguard-rules.txt
@@ -1,4 +1,5 @@
--keep class com.microsoft.** { *; }
+# Keep the names of JSON-serializable objects.
+-keepnames public class * implements com.microsoft.graph.serializer.IJsonBackedObject
 
 ## GSON 2.2.4 specific rules ##
 # Source: https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg


baywet (approved):
thanks a lot for the contribution. Would you mind replicating the change onto https://github.com/microsoftgraph/msgraph-beta-sdk-java ?

Digipom:
Hi @baywet, looks like it was already replicated! Thanks for merging.

PR #1062, opened 2022-04-27
Update WorkbookWorksheetRangeRequestBuilder.java

BaranwalG:
For specified range this API is failing, need this change to make API work for specified change also.

Fixes # Issue : #1063 Specified range is not being pull in worksheet in graph api.

- This change requires to support specified range of worksheet.

### Other links

-


ramsessanchez:
Hi @BaranwalG, unfortunately I cannot approve this PR. This SDK library is auto generated weekly and thus any changes made directly to any classes will be overwritten. The generation is based on a separate metadata file with the API entities and their corresponding functionality. I will check with the maintainer of the specific API that you are using and see if there are any changes that they can make on their side or if there is a suggested approach that would yield your desired result. Thanks for using our SDK and I'll be sure to follow up if there are any changes to expect in a later release that would help your issue.

PR #1080, opened 2022-05-26
Bump graph core to 2.0.13

ramsessanchez:
Bumping the dependency on graph core to 2.0.13


baywet (changes requested):
good initiative here, no need to make a manual release however.
dependabot will pick it up tomorrow, and we can release that change as part of the next weekly generation.

PR #1146, opened 2022-08-17

ramsessanchez:

 The commit hash is _2a8956a94940ad33e847b8a7616f3319d6b24331_. 


baywet (changes requested):
@ramsessanchez to investigate why this one is failing

baywet (changes requested):
I think we should close this one, it shows no actual changes, which is a bit weird considering we had a breaking change. (maybe a regression from another fix?)

ramsessanchez:
@baywet yeah, this is strange, Im having trouble running pkg diff locally to verify, but as far as I can tell from the github comparison I only notice namespace and comment changes.

ramsessanchez:
@baywet, Actually I know what happened. Last week with the generation fix I did work to remove a lot of unnecessary import statements. I believe I deleted a few import statements that belonged in different namespaces so this time around it had trouble finding those properties. Todays changes made all properties fully qualified so this should no longer be an issue going forward since a fully qualified namespace removes the need for import statements.

PR #1147, opened 2022-08-17
GraphServiceClient

ramsessanchez:
Adding GraphServiceClient class for use in preview. 
Removing old tests. 
Updating dependencies and versions for use in preview.


gradle/dependencies.gradle
```diff
@@ -6,10 +6,18 @@ dependencies {
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.2'
     testImplementation 'org.mockito:mockito-inline:4.5.1'
     
-    api 'com.google.code.gson:gson:2.9.0'
+    implementation 'com.google.code.gson:gson:2.9.0'
```
baywet, line 9: do we need to have a dependency on gson here anymore? I believe it's going to be brought by the serialization package.

gradle/dependencies.gradle
```diff
@@ -6,10 +6,18 @@ dependencies {
...
     
     implementation 'com.squareup.okhttp3:okhttp:4.9.3'
+    compileOnly 'com.github.spotbugs:spotbugs-annotations:4.7.0'
```
baywet, line 12: same question about having a direct dependency on okhttp

src/main/java/com/microsoft/info/Constants.java
```diff
@@ -0,0 +1,22 @@
...
+    /** The application ID to use for unit testing */
+    public static final String APPID = "app-id";
+    /** The user email to use for unit testing */
+    public static final String USERNAME = "user@email.com";
+    /** The user password to use for unit testing */
+    public static final String PASSWORD = "password";
+    /** The tenant ID to use for unit testing */
+    public static final String TENANTID = "tenantid";
+    /** The client secret to use for unit testing */
+    public static final String CLIENTSECRET = "clientsecret";
```
baywet, line 19:
```suggestion
```

I don't believe we need those constants anymore since the tests have been removed

src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
```diff
@@ -0,0 +1,79 @@
...
+import com.microsoft.kiota.authentication.AzureIdentityAuthenticationProvider;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+import okhttp3.OkHttpClient;
+
+import javax.annotation.Nonnull;
+
+public class GraphServiceClient extends microsoft.graph.BaseGraphServiceClient implements IBaseClient {
+
+    private RequestAdapter requestAdapter;
```
baywet, line 20: I don't think the field should be defined again here, we should refer directly to the field in the class that's generated.

baywet (changes requested):

src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
ramsessanchez: Unfortunately that would mean that we could not change the RequestAdapter since we do not have get/set methods in the generated class, hence the need for this field here. Unless the idea is that we only change the RequestAdapter when we create a new GraphServiceClient using 'new GraphServiceClient(requestAdapter)' which would require a new instantiation of the GraphServiceClient. An alternative would be to simply add the get/set methods to the generated class by implementing IBaseClient in BaseGraphServiceClient and thus remove the need for GraphServiceClient to implement IBaseClient and instead only require the extension of BaseGraphServiceClient.

baywet: right because it's private for now. And @andrueastman had the [same problem with dotnet](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/a6059f734d959a454e5017a0372d5feacbc238dd/src/Microsoft.Graph/GraphServiceClient.cs#L73)
What would you think of a common refiner in Kiota to switch this property to public just for the API client class (not the other request builders) or add accessors?

gradle/dependencies.gradle
ramsessanchez: correct

baywet:
```suggestion
```

src/main/java/com/microsoft/info/Constants.java
ramsessanchez: I thought it would be good to keep them incase we need them for future tests

baywet: I don't think we should have any unit or integration test at the service library level.
We have unit tests in our conversion lib and generator to cover the accuracy of the generation.
We have integration tests at the raptor level.
But manually testing generated code is counter productive IMHO.

src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
ramsessanchez: I think accessors would be a better method, it would make it obvious that we intend to grant the user the ability to get/set them in the client class. Do we want to add the implements IBaseClient to the api client class? Or would adding that mean having a direct dependency on core which would unnecessary for other generated API's?

baywet: You can also add the accessors to the IBaseClient if you need to access the request adapter in core, which you'll need for large file upload and batch support.
As long as they match what we generate, the resulting api client will satisfy the interface. They won't need to be defined in this hand crafted client anymore however since they'll come from the generated class.

ramsessanchez: https://github.com/microsoftgraph/msgraph-sdk-java-core/blob/v3/telemetryRefactor/src/main/java/com/microsoft/graph/Requests/IBaseClient.java
The IBaseClient really only contains the accessors, I say we just add them to the generated class/implement the interface and we should be good to go

ramsessanchez: I'll delete the accessors and the RequestAdapter property from this GraphServiceClient class

baywet (approved):

PR #1155, opened 2022-08-25
[ItemAnalytics] - Item Analytics Builder

Jatish-Khanna:
As per the new documentation, to fetch item analytics, it should add either lastSevenDays or allTime in the URL segment. The methods are available ItemAnalyticsRequestBuilder class.

GET /drives/{drive-id}/items/{item-id}/analytics/allTime
GET /sites/{site-id}/analytics/allTime
GET /sites/{site-id}/lists/{list-id}/items/{item-id}/analytics/allTime
GET /drives/{drive-id}/items/{item-id}/analytics/lastSevenDays
GET /sites/{site-id}/analytics/lastSevenDays
GET /sites/{site-id}/lists/{list-id}/items/{item-id}/analytics/lastSevenDays

So the final Java snipped would be working and there will be proper response from Microsoft Graph API -

GraphServiceClient graphClient = GraphServiceClient.builder().authenticationProvider( authProvider ).buildClient();

ItemActivityStat itemActivityStat = graphClient.drives("{drive-id}").items("{item-id}").analytics().allTime()
	.buildRequest()
	.get();

Fixes #

-

### Other links

-

### DIFF
--- a/src/main/java/com/microsoft/graph/requests/DriveItemRequestBuilder.java
+++ b/src/main/java/com/microsoft/graph/requests/DriveItemRequestBuilder.java
@@ -124,11 +124,11 @@ public class DriveItemRequestBuilder extends BaseRequestBuilder<DriveItem> {
     /**
      * Gets the request builder for ItemAnalytics
      *
-     * @return the ItemAnalyticsWithReferenceRequestBuilder instance
+     * @return the ItemAnalyticsRequestBuilder instance
      */
     @Nonnull
-    public com.microsoft.graph.requests.ItemAnalyticsWithReferenceRequestBuilder analytics() {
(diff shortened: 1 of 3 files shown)


baywet (changes requested):
Hi @Jatish-Khanna 
Thanks for reporting this issue and for the contribution.
Unfortunately we cannot accept it because this section of the code is automatically generated and would get replaced at the next weekly generation.
The reason why the paths are not present is because the metadata is incorrect.
The itemAnalytics properties for site and drive item should have a ContainsTarget="true" attribute.
https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsv1.0.xml
We do have a transform script to "patch" the metadata if you feel like contributing over there.
In the meantime I'll open an issue in the metadata repo and engage with the concerned product team.

PR #1172, opened 2022-09-07
Remove implementations for Kiota Components

ramsessanchez:
Kiota dependencies in core have been upgraded to 'api' meaning they are now visible/usable here. We can remove the Kiota components from here and use them through transitivity.


baywet (changes requested):
I'm not sure what prompted this change. The service library should have direct dependencies on abstractions, serialization, and authentification at least because it directly references classes from there

PR #1185, opened 2022-09-16
Source/Target compatibility to 8

ramsessanchez:
Target and source compatibility to version 8. 
Approve once core compatibility is also at 8: https://github.com/microsoftgraph/msgraph-sdk-java-core/pull/622

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,8 +154,8 @@ def fixAscNames = { name ->
 
 compileJava {
   options.compilerArgs << "-parameters"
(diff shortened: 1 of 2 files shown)


baywet (changes requested):
Version of core needs bumping

ramsessanchez:
@baywet this can be force pushed

baywet:
now that core is published, we're pending on this + a generation. https://github.com/microsoft/kiota/pull/1849

PR #1226, opened 2022-10-17
Kiota/v1.0/pipelinebuild/88901

ramsessanchez:
@baywet This uses the recently fixed long discriminator method and runs linting on the MacOs github runner. 
Long discriminator fix is good, will re-run this with the 1es pool and verify that all is set. If that passes, we should be good to go with tomorrow's generation update.


ramsessanchez:
Using 1ES runner for linting

ramsessanchez:
superseded by #1234

PR #1243, opened 2022-10-25
Generated v1.0 models and request builders using Typewriter

ramsessanchez:
This pull request was automatically created by the GitHub Action, **create pull request**. 

 The commit hash is _6160ea38441292ba209c6d29570d96eb8b0962d9_. 

 **Important** Check for unexpected deletions or changes in this PR.

Issue #1063, opened 2022-04-27
Specified range is not being pull in worksheet in graph api.


BaranwalG:
### Expected behavior

API /me/drive/items/{id}/workbook/worksheets/{id|name}/range(address='<address>') should give range for given address.

### Actual behavior

API /me/drive/items/{id}/workbook/worksheets/{id|name}/range(address='<address>') giving full range of worksheet.

### Steps to reproduce the behavior
execute API: /me/drive/items/{id}/workbook/worksheets/{id|name}/range(address='<address>') with proper input in java sdk.

ramsessanchez:
https://github.com/microsoftgraph/msgraph-sdk-java/pull/1062#issuecomment-1112630709

jkscx:
Hello, we are facing the same issue. Namely, that the "GetRange" endpoint doesn't respect the given address/range.  

The endpoint works normally when used in Graph Explorer or with plain HTTP requests in Java, but when we use the GraphAPI-Java-SDK in the same way, it doesn't work. Or to be more precise, it does the same action as if a blank string "" was entered for the address of the range.

Also, we can confirm that there exists a bug in the original GraphAPI-Java-SDK code for which @BaranwalG submitted a fix. Namely, without his fix it misses to actually use the address field of the range and that's why each GetRange request has the same response as if one didn't enter any address for it. 

Java code which can be used to see that providing an address for the GetRange endpoint doesn't do anything in the GraphAPI-Java-SDK, i.e. the address gets ignored:

         var valueOfA1 = _graphApiClient
                 .drives(DRIVE_ID)
                 .items(ITEM_ID)
                 .workbook()
                 .worksheets("Sheet1")
                 .range(WorkbookWorksheetRangeParameterSet.newBuilder()
                         .withAddress("A1")
                         .build())
                 .buildRequest()
                 .get();
         LOGGER.info("Address should be 'Sheet1!A1', but is actually:\n{}\n", valueOfA1.address);

But the same request as a plain HTTP request works, i.e. returns the desired range and has the correct address:

         GET https://graph.microsoft.com/v1.0/drives/{drive-id}/items/{item-id}/workbook/worksheets/Sheet1/range(address='A1')

Hope you can help us with this.

BaranwalG:
@baywet any update on this!

JimeFuture:
Is there any news on this topic? We are also missing the row and the column function for ranges.

ramsessanchez:
[...]

PR #1245, opened 2022-10-26
Update generated files with build 90185

ramsessanchez:
Superseding #1243 using  https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/857
This addresses https://github.com/microsoftgraph/msgraph-sdk-java/issues/1063


ramsessanchez:
Still needs changelog

baywet:
@ramsessanchez good catch! This is most likely caused by [this change in the metadata](https://github.com/microsoftgraph/msgraph-metadata/pull/213). Reports are a bit all over the place metadata wise, let me confirm this is expected.

baywet:
After looking into it further it seems that all the getXXX reports make sense for that change. However, there are a few reports that are still on the older format (all the ones not starting with get). Trying to get a confirmation on whether these will align too from @darrelmiller . https://learn.microsoft.com/en-us/graph/api/intune-deviceconfig-reportroot-deviceconfigurationuseractivity?view=graph-rest-1.0
@irvinesunday do you think we could tweak the filter to only include what starts with "get" if those reports are not going to align on the service side?

baywet:
see this article as well https://devblogs.microsoft.com/microsoft365dev/changes-to-microsoft-365-apps-usage-reports-api-in-microsoft-graph/

irvinesunday:
> After looking into it further it seems that all the getXXX reports make sense for that change. However, there are a few reports that are still on the older format (all the ones not starting with get). Trying to get a confirmation on whether these will align too from @darrelmiller . https://learn.microsoft.com/en-us/graph/api/intune-deviceconfig-reportroot-deviceconfigurationuseractivity?view=graph-rest-1.0 @irvinesunday do you think we could tweak the filter to only include what starts with "get" if those reports are not going to align on the service side?

Created an issue to track this: https://github.com/microsoftgraph/msgraph-metadata/issues/216

baywet:
Thanks. Bottom line I don't think we should release this version since it breaks things that are working today on top of fixing broken things. Thoughts @darrelmiller @andrueastman (and everyone on this thread)?

irvinesunday:
Here's a PR resolving that @baywet: https://github.com/microsoftgraph/msgraph-metadata/pull/217

baywet:
thanks, closing this one, we'll trigger a new generation.

PR #1254, opened 2022-11-08
Update generated files with build 91765

ramsessanchez:
Fixes #

-

### Other links

-


baywet:
@ramsessanchez I believe this one was a misfire from the tests for beta, can you confirm and close/delete?

ramsessanchez:
@baywet This is a typical v1 update for the kiota generated library. This is valid and should be pushed. V1 doesn't suffer from the long name issues found in beta since there aren't any excessively long classes generated.

baywet (approved):
thanks for clarifying!

PR #1283, opened 2022-12-13

ramsessanchez:

 The commit hash is _b2053486edbed8520fbc142a04e0e2e74a8298bb_. 


baywet (approved):
@ramsessanchez to update the changelog
@andrueastman it seems the InformationProtection entity type has been changed to a complex type. Which is technically a binary breaking change. However since it's only used as a singleton for the identity protection root, I don't think it matters (we can consider it a fix), nobody could have used it anyway. Thoughts?

andrueastman (dismissed):
> it seems the InformationProtection entity type has been changed to a complex type. Which is technically a binary breaking change. However since it's only used as a singleton for the identity protection root, I don't think it matters (we can consider it a fix), nobody could have used it anyway. Thoughts?

Hey @baywet, I agree with this. 
The type could not really have been used for payload serailizing/deserializing but for simply creating the navigation path for the singleton. 

I also found this which suggests that this is a fix and to help better organize the API
https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Workloads/pullrequest/7263167

zengin (approved):

PR #1357, opened 2023-02-28

ramsessanchez:

 The commit hash is _c56d50c2f70708aa222f6b77cca0599490113f29_. 


src/main/java/com/microsoft/graph/models/AppCredentialRestrictionType.java
```diff
@@ -7,40 +7,36 @@
 
 
 /**
- * The Enum Lobby Bypass Scope.
```
andrueastman, line 10: Just to follow up here.
The fix was applied via https://msazure.visualstudio.com/One/_git/AD-AggregatorService-Workloads/pullrequest/7663994?_a=files

Fix can be confirmed on staging metadata. But is yet to propagate fully on prod. ETA is latest tomorrow. 

cc @ramsessanchez @baywet @Ndiritu @SilasKenneth

baywet:
closing to get the fix for the breaking change.

PR #1371, opened 2023-03-08
Generated  models and request builders

andrueastman:
This pull request was automatically created by Azure Pipelines. **Important** Check for unexpected deletions or changes in this PR.


andrueastman:
@baywet I believe the erroneous removal of the `validateProperties` should be fixed here.

Otherwise, only the removal of the `memberOf` casts and `skip` method are the only expected deletions.

baywet (approved):

PR #1438, opened 2023-05-04
Update Graph Service Client to use IbaseClient

ramsessanchez:
Implements IBaseClient, this is in line with dotNet and will help going forward with batch requests.

### DIFF
--- a/src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
+++ b/src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
@@ -48,6 +48,7 @@ public class GraphServiceClient extends com.microsoft.graph.BaseGraphServiceClie
      * @param authenticationProvider The AuthenticationProvider for this GraphServiceClient.
      * @param client The OkHttpClient for the GraphServiceClient.
      */
+    @SuppressWarnings("LambdaLast")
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/serviceClient/GraphServiceClient.java
```diff
@@ -3,19 +3,21 @@
...
 import com.microsoft.graph.requests.GraphClientOption;
+import com.microsoft.graph.requests.IBaseClient;
 import com.microsoft.info.Constants;
 import com.microsoft.kiota.RequestAdapter;
 import com.microsoft.kiota.authentication.AnonymousAuthenticationProvider;
 import com.microsoft.kiota.authentication.AuthenticationProvider;
 import com.microsoft.kiota.authentication.AzureIdentityAuthenticationProvider;
 
 import okhttp3.OkHttpClient;
+import org.jetbrains.annotations.NotNull;
```
MIchaelMainer, line 14: Why do we use `org.jetbrains.annotations.NotNull`?

ramsessanchez: ah, good catch, fixed

MIchaelMainer (approved):

PR #1699, opened 2023-12-12
Update generated files with build 130458 [skip ci]

ramsessanchez:
Fixes #

-

### Other links

-


baywet (changes requested):
@ramsessanchez I'm a bit confused by this latest generation
e.g. 
src/main/java/com/microsoft/graph/generated/admin/serviceannouncement/healthoverviews/item/issues/IssuesRequestBuilder.java
was moved to 
src/main/java/com/microsoft/graph/generated/com/microsoft/graph/admin/serviceannouncement/healthoverviews/item/issues/IssuesRequestBuilder.java

We're clearly repeating the com/microsoft/graph section. I think this has to do with how kiota does path reduction for Java: Since the path doesn't end with com/microsoft/graph and since it's the root namespace, it gets added to map to Java file structure.

My suggestion here:
1. revert the output path during the generation so we don't get this side effect.
2. add a script after generation in the pipeline so the content gets moved to the generated sub-directory.

I know this is a bit hacky but I don't think there's a better way around without severely impacting kiota users.

ramsessanchez:
@baywet I didnt update the pr, this should not be merged. Closing.

PR #1749, opened 2024-01-20
Create upgrade-to-v6.md

ramsessanchez:
V6 upgrade guide

### DIFF
--- a/docs/upgrade-to-v6.md
+++ b/docs/upgrade-to-v6.md
@@ -1,0 +1,507 @@
+# Microsoft Graph Java SDK v6 Changelog and Upgrade Guide 
+
+This document provides a list of changes and upgrade considerations for the Microsoft Graph Java SDK v6 release.
+
+## Overview 
+
+Version 6.0.0 of the Microsoft Graph Java SDK is based on the new [Kiota](https://github.com/microsoft/kiota) code generation tool. By using Kiota the SDK is now able to support a broader range of Microsoft Graph API enpoints while also being more intuitive. Furthermore, [Graph-Core](https://github.com/microsoftgraph/msgraph-sdk-java-core) has also been updated to v3 which provides a great amount of added functionality; this makes v6 of the SDK a significant upgrade from v5.
+
+## Breaking Changes
+
+The following is a list of breaking changes that users upgrading will want to consider while upgrading to v6 of the SDK.
+
+### Namespace Changes and Disambiguation
+
+Historically the SDK as well as the underlying Graph-Core library have all used the same parent namespace, `com.microsoft.graph.*`. This has caused issues for users who have wanted to use both the [v1.0](https://github.com/microsoftgraph/msgraph-sdk-java), [beta](https://github.com/microsoftgraph/msgraph-beta-sdk-java) and [core](https://github.com/microsoftgraph/msgraph-sdk-java-core) libraries in the same project. 
+This has the following implications:
+- The v1.0 version of the SDK maintains the `com.microsoft.graph.*` namespace.
+- The beta version of the SDK is now disambiguated to `com.microsoft.graph.beta.*`.
+- The core library is now disambiguated to `com.microsoft.graph.core.*`.
+- Models types are now stored in `com.microsoft.graph.models`/`com.microsoft.graph.beta.models` for v1.0 and beta respectively.
+- RequestBuilder and RequestBody types reside in namespaces relative to the path they are calling. e.g. The `SendMailRequestBuilder` and `SendMailPostRequestBody` types will reside in the `com.microsoft.graph.users.item.sendmail`/`com.microsoft.graph.beta.users.item.sendMail` namespace if you are sending mail via `graphClient.me().sendMail().post(sendMailPostRequestBody)`. 
+
+### Authentication 
+
+The `GraphServiceClient` class now accepts an instance of `TokenCredential` from AzureIdentity directly rather than an instance of `TokenCredentialAuthProvider`.
+Thus 
+```java 
+final DeviceCodeCredential credential = new DeviceCodeCredentialBuilder()
+                .clientId(appId)
+                .challengeConsumer(challenge -> System.out.println(challenge.getMessage()))
+                .build();
+final TokenCredentialAuthProvider tokenCredential = new TokenCredentialAuthProvider(scopes, credential);                
+GraphServiceClient graphClient = GraphServiceClient.builder()
+    .authenticationProvider(tokenCredential)
+    .buildClient();
+```
+becomes
+```java
+final DeviceCodeCredential credential = new DeviceCodeCredentialBuilder()
+                .clientId(appId)
+                .challengeConsumer(challenge -> System.out.println(challenge.getMessage()))
+                .build();
+GraphServiceClient graphClient = GraphServiceClient(credential, scopes);
+```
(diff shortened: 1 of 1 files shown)


docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,507 @@
...
+In v6 the same call would look like the following:
+```java
+Message singleMessage = graphClient.me().messages().byMessageId("<Message Id>").get();
+```
+
+### Option Class Removal
+
+The `Option` class has been removed and is no longer used to define query parameters, headers, or function parameters. These classes have been replaced with more specific and intuitive implementations. 
+
+### Headers
```
andrueastman, line 158: Maybe the `Headers` and the `Query Parameter` sections be subsections under `Option Class Removal`?

andrueastman (commented):
Looks great really @ramsessanchez 

- Should the PR be targeting the dev or the kiota branch
- Any chance we can also add a section in `Readme.md` linking to this file in the event one reads they can discover this as well?

docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,507 @@
...
+try {
+    User user = graphClient.me().get();
+} catch (ApiException e) {
+    System.out.println(e.getMessage());
+}
+```
+
+### Drive Item Paths
+
+The current CSDL to OpenAPI conversion process avoids generation of redundant paths, which impacts request builders for driveItems. To mitigate this paths should be available through alternative paths as documented in the reference documentation as seen [here](https://learn.microsoft.com/en-us/graph/api/driveitem-list-children?view=graph-rest-1.0&tabs=http#http-request).
```
baywet, line 244:
```suggestion
The SDK generation process avoids generation of redundant paths, which impacts request builders for driveItems. To mitigate this paths should be available through alternative paths as documented in the reference documentation as seen [here](https://learn.microsoft.com/en-us/graph/api/driveitem-list-children?view=graph-rest-1.0&tabs=http#http-request). 
```

(I believe most people reading this won't have an understanding of our complex pipeline)

docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,507 @@
...
+```
+
+#### Upload a Small File with conflictBehavior Set
+
+To upload a small file (size should not exceed 4mb according to the [docs](https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0&tabs=http)) and set the `conflictBehavior` [instance attribute](https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0#instance-attributes) you'll need to do it this way:
+
+```java
+RequestInformation requestInformation = graphClient.drives().byDriveId("Drive-Id").items().byDriveItemId("root:/MediaMeta.xml").content().toPutRequestInformation(file);
+// 'file' should be an inputStream
+URI uriIncludesConflictBehavior = new URI(requestInformation.getUri().toString()+"?@microsoft.graph.conflictBehavior=rename");
```
baywet, line 295: can you create an issue in the metadata repo if one doesn't exist already so we add an additional annotation where this is supported and this query parameter is emitted please?

docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,507 @@
...
+OData allows downcasting the result to the right specialized type, meaning that users can have access to all properties in the response, rather than pulling those out from the AdditionalData holder. The fluent API pattern now comes with enriched segments allowing users to request a specific Odata type in the event that an endpoint supports multiple types. 
+
+For example:
+```java
+// Get the collection of Users in a group
+// API Endpoint: /groups/{id}/members/microsoft.graph.user
+UserCollectionResponse users = graphClient.groups().byGroupId("Group-Id").members().graphUser().get();
+// Get the collection of Applications in a group
+// API Endpoint: /groups/{id}/members/microsoft.graph.application
+ApplicationCollectionResponse applications = graphClient.groups().byGroupId("Group-Id").members().graphApplication().get();
```
baywet, line 506: the raw count is missing

docs/upgrade-to-v6.md
```diff
@@ -0,0 +1,507 @@
+# Microsoft Graph Java SDK v6 Changelog and Upgrade Guide 
+
+This document provides a list of changes and upgrade considerations for the Microsoft Graph Java SDK v6 release.
+
+## Overview 
+
+Version 6.0.0 of the Microsoft Graph Java SDK is based on the new [Kiota](https://github.com/microsoft/kiota) code generation tool. By using Kiota the SDK is now able to support a broader range of Microsoft Graph API enpoints while also being more intuitive. Furthermore, [Graph-Core](https://github.com/microsoftgraph/msgraph-sdk-java-core) has also been updated to v3 which provides a great amount of added functionality; this makes v6 of the SDK a significant upgrade from v5.
```
baywet, line 7: A table of content might help here.

baywet (commented):
This is great work! I love the fact that it showcases more capabilities with less code in general!
@maisarissi can you schedule a slot on the pnp call post GA so Ramses and you can demonstrate all that goodness and amplify the roll out?

ramsessanchez:
> * Should the PR be targeting the dev or the kiota branch

I wasn't sure here. My thinking was that once we GA this branch will hold all the 6.0 stuff so putting this here would be okay since it doesn't affect functionality.

docs/upgrade-to-v6.md
ramsessanchez: Do you mean: 
graphClient.groups().byGroupId("").members().get()?

baywet: no I meant the new endpoints we support like `graphClient.groups().count()`. This SDK didn't use to support it.

docs/upgrade-to-v6.md
ramsessanchez: The conflictBehavior option is referred to as an 'InstanceAttribute' and not an QueryParameter in the documentation, this could be why it is not emitted as a query parameter. Would this set the precedent that these instance attributes should be treated as QueryParams and thus belong in the query parameter options list? Would this be across the languages? And would this be more of a decision for Kiota or only for our graph repo?

docs/upgrade-to-v6.md
ramsessanchez: The SDK as of v5 has supported count requests. 
```graphClient.groups().count().buildRequest();``` this is a valid request in v5

docs/upgrade-to-v6.md
baywet: Kiota doesn't know anything about OData conventions.
This would be for openapi.net.odata and for the metadata repos (I suggest we start in the metadata repo so we can identify the right annotation).
I'm not aware of a way to document instance annotations in CSDL today but it might be worth exploring.
As for the query parameter itself, we already have a way to document those through annotations.
Start a discussion there, we'll eventually figure it out

docs/upgrade-to-v6.md
ramsessanchez: I think the surface area of where .count() is supported has expanded, but I dont believe .count() is a new new feature. I can add a note to explain that the support has increased which I agree is important. Where do you think would be a good place to add the note or even section?

baywet: Really? I thought we didn't have them for dotnet/java with typewriter. Ok disregard this comment then

docs/upgrade-to-v6.md
andrueastman: Ramses has created the issue at https://github.com/microsoftgraph/msgraph-metadata/issues/560

andrueastman:
> My thinking was that once we GA this branch will hold all the 6.0 stuff so putting this here would be okay since it doesn't affect functionality.

That's valid. In my head, the reasoning is also similar in that the kiota branch will go into the branch eventually so I thought it may make sense to have the guide initially live with the 6.0 stuff and they go in together.

baywet (approved):

PR #1799, opened 2024-02-12
clean up workflows

### LINKED ISSUE #1782: Clean up long term branches after 6.1.0 Release
Clean up refs to feature/6.0 in Dependabot scripts, GitHub Actions, Policies, and finally delete the branch.

ramsessanchez:
feature/6.0 has now been merged into Dev.
This means the branch can be safely deleted and workflow pipelines and refs can also be deleted.
Fixes #1782


.github/dependabot.yml
```diff
@@ -39,7 +39,6 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
-  target-branch: feature/6.0
```
andrueastman, line 42: Should we be removing the entire section since it already exists above (line 3)

.github/workflows/preview-and-release.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   maven_Preview:
```
andrueastman, line 19: QQ: Do we still need to publish to the preview feed? Or should we drop the entire section too?

.github/workflows/gradle-build.yml
```diff
@@ -21,8 +21,8 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          java-version: 18
-          distribution: 'adopt'
+          java-version: 20
```
baywet, line 24:
```suggestion
          java-version: 21
```
21 is an Lts, 20 isn't

.github/workflows/preview-and-release.yml
baywet: Even though I don't think many people use the snapshot feed, I still think having the ability to preview things is interesting, especially since the only cost is to maintain this part of the workflow.
We might reevaluate when migrating to ESRP release

baywet (changes requested):

baywet (approved):

baywet (approved):

PR #1903, opened 2024-03-26
working elsewhere is been identified as 0 not 4!

DenWin:
Fixes : The description is off

- I do not know, whether it is intentionally or a bug, but if one requests getschedule an appointmeint "shownAs" as "working elsewhere" is represented with a zero, not 4.

### Other links

-


DenWin:
I checked [graph.microsoft.com/v1.0/$metadata](https://graph.microsoft.com/v1.0/$metadata)
In there workingElsewhere is indeed 4, thought it is not shown accordingly in availabilityView if one requests "getschedule"

Issue #1978, opened 2024-05-02
Update release workflow to trigger on tags for maven central releases


baywet:
The current workflow triggers on push to dev
https://github.com/microsoftgraph/msgraph-sdk-java/blob/a460c841ca789d3b3eb513d14c3da4a44e34e350/.github/workflows/preview-and-release.yml#L5
https://github.com/microsoftgraph/msgraph-sdk-java/blob/a460c841ca789d3b3eb513d14c3da4a44e34e350/.github/workflows/preview-and-release.yml#L52

This is problematic because it creates more releases than we want.

TODO:
- [ ] remove the trigger for the support branch (keep dev for snapshots)
- [ ] add a trigger on tags v*
- [ ] update the filter for central releases to check for tags only
- [ ] remove approval on the maven central environment
- [ ] replicate on beta

baywet:
To recap what was said in teams and during meetings: 
Before we can remove approval, we need to make sure releases are not going to go out inadvertly.
I suggested we add a filter to the maven central release job so we only consider tags created by the release please bot moving forward.
Once this is done, we can close this issue, and remove the approval.

PR #2005, opened 2024-05-20
Trigger maven releases on tag creation

Ndiritu:
part of https://github.com/microsoftgraph/msgraph-sdk-java/issues/1978


baywet (approved):
We should take this opportunity to align branching on main as well.

PR #2017, opened 2024-05-23
Merge GitHub release steps to maven release workflow

Ndiritu:
Uploading the build artifacts (JARs) for 6.11.0's GitHub release [fails ](https://github.com/microsoftgraph/msgraph-sdk-java/actions/runs/9204821667) since it can't find the build artifacts to upload to the GitHub release.

The workflow was expecting build artifacts to be produced by the maven release step, however this step was recently updated to upload build artifacts [on tag creation](https://github.com/microsoftgraph/msgraph-sdk-java/commit/1a87cfd0f86c151032dbb2440ff496d803821491#diff-207e26d4688ea5c2f4db22c21b9ed04a15fc08ff9bc1086a535461559f825a29R54)

This PR adds the GitHub release step after a maven release so that build artifacts generated in the maven release are immediately available for the GitHub release.


.github/workflows/gradle-build.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         with:
           workflow: preview-and-release.yml
           workflow_conclusion: success
-          branch: main
+          branch: dev
```
andrueastman, line 75: I'm not sure this will work. The branch no longer exists. Yes?

baywet: We deleted the branch as part of branching alignment. 
The historic workflow runs are still present with assets though.
@Ndiritu I don't think we need to make that change. This check is optional

Ndiritu: It's optional but the "Git release" workflow which actually uploads build artifacts to the release created by release-please fails because it's looking for a previous successful build workflow run. The build workflow is unsuccessful because one step fails (compare packages)

baywet: I think there might be a misunderstanding here, let me try to bring additional context.

This job (compare-packages) is a CI job, that's not enforced as a gate check. Yes it uses the same taks/branching mechanism as the git release job, but it's orthogonal. And it's ok to have it fail the first time because the upload artifact task it depends on, hasn't run on that branch before. It should resolve itself over time.

[That job](https://github.com/microsoftgraph/msgraph-sdk-java/blob/62019c4eab96c84eede8bb483193bda203f27588/.github/workflows/git-release.yml#L10) (git release) depends on the same upload to successfully create the release, and has a dependency on a branch. Since everything now triggers on tags, I suggest we merge that workflow with the one in the next paragraph. (we'd be able to use the standard download artifact task, no more dependency on a specific branch, etc)

[This other job](https://github.com/microsoftgraph/msgraph-sdk-java/blob/62019c4eab96c84eede8bb483193bda203f27588/.github/workflows/preview-and-release.yml#L53) (maven release) is the one uploading the artefact both previous jobs depend on. Again, I'm suggesting we bring the git release job as a successor to the maven release one, simplifying everything in the process.

Ndiritu: This makes sense @baywet. I mis-understood where the artifacts were being pulled from. Updated the PR description & title

baywet (approved):

Issue #2078, opened 2024-07-11
JavaDoc jar shipped with the SDK is empty


Ndiritu:
### Describe the bug

Empty JavaDoc JAR is hinders set-up time with the SDK.

What's missing from this upgrade guide is a link to _any_ v6 JavaDocs. The javadoc jar on [Maven Central](https://repo1.maven.org/maven2/com/microsoft/graph/microsoft-graph/6.13.0/) is empty, nothing but a `META-INF` directory. Ditto for _all_ the other v6 versions I checked. I'm trying to upgrade a v5 app and frustrated by the lack of JavaDocs. In particular, I'm stumbling around trying to find the `import` and API for classes like `PageIterator`, `RequestConfiguration `, etc. If you won't provide JavaDocs, then at least show the `import`s in your examples.

_Originally posted by @johnthad in https://github.com/microsoftgraph/msgraph-sdk-java/issues/2070#issuecomment-2209492155_
            

### Expected behavior

JavaDoc Jar to contain relevant docs

### How to reproduce

,

### SDK Version

6.x

### Debug output

<details><summary>Click to expand log</summary>
```

<Log output here>

```
</details>

dias717:
@Ndiritu,

I agree with you. The JavaDoc does not work in Intellij. The reason is that the *-javadoc.jar file has only one manifest file in it. The other libraries in the *-javadoc.jar archive contain many files and weigh several megabytes. If I do the following **workaround**, then JavaDoc works in Intellij. Microsoft should do something similar and provide us with a working JavaDoc!


WORKAROUND TO USE THE JAVA DOC FOR MS GRAPH 6 in Intellij on Windows 10


**GENERATE JAVA-DOC:**

```
javadoc -d /c/Users/MY-USER-NAME/microsoft-graph-docs \
-sourcepath /c/Users/MY-USER-NAME/microsoft-graph-sources \
-subpackages com.microsoft.graph \
[...]

Issue #1893, opened 2024-03-20
JavaDoc has to be modified manually to map to its corresponding classes


coinzz:
### Expected behavior
JavDoc should be readable in IDE simply after downloading it.

### Actual behavior
JavDoc package structure has to be modified manually, since the paths do not match:
Code package path: `com.microsoft.graph.*`
JavaDoc package path: `com.microsoft.graph.generated.*`

After moving all contents in the sources jar from generated one folder up, the IDE is able to map the JavaDoc to it's corresponding classes.

### Steps to reproduce the behavior
Download current version of graph [6.4.0](https://mvnrepository.com/artifact/com.microsoft.graph/microsoft-graph/6.4.0) and its corresponding sources.

baywet:
@coinzz Thanks for using the java SDK and for reporting this
@ramsessanchez FYI this is why I was a bit uncomfortable when you suggested moving the generated stuff to a dedicated directory without updating the package name. Couldn't remember the rationale, but I somehow remembered it'd break something.
@Ndiritu whenever you have some time, here are the steps to take:
1. update the weekly generation cleanup script to delete what's in `src/main/java/com/microsoft/graph` excluding `info` and `serviceclient`.
2. update the weekly generation path to stop appending a "generated" segment.
3. kick-off a generation.
4. make sure the generated directory is deleted from PR.

(let me know if you have questions)

poschi3:
The problem still exists. Without JavaDoc and sources it is really hard to work with the SDK. The problem still exists in version 6.13.0.

![image](https://github.com/user-attachments/assets/355f5557-a978-40d3-8890-3e0984cd5fb6)

faisalferoz:
Issue mentioned by @poschi3 still exists in version 6-25.1

PR #2298, opened 2025-02-12
feat: add ChangeNotification models

Ndiritu:
depends on https://github.com/microsoftgraph/msgraph-sdk-java-core/pull/1837

closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/1891

PR #2317, opened 2025-02-27
fix: resolve javadoc and source locations

andrueastman:
fixes https://github.com/microsoftgraph/msgraph-sdk-java/issues/2078
fixes https://github.com/microsoftgraph/msgraph-sdk-java/issues/1893

This PR includes the following changes.
- enables the javadoc task to ensure that javadocs are part of release task artifacts
- adds a step in CI/CD stages to move sources in the `/generated` folder up one folder so that the generated sources are in the correct package structure

Considerations.
As much as updating the generation pipeline to place the generated files in the correct package structure is a good idea, we have a number of hand writter files in the root path such as https://github.com/microsoftgraph/msgraph-sdk-java/pull/2298 that would be overwritten or replaced. The alternative would involve keeping a record of handwritten files and moving them just in time but that would increase the maintenance burden.

### DIFF
--- a/.github/workflows/api-level-lint.yml
+++ b/.github/workflows/api-level-lint.yml
@@ -17,6 +17,9 @@ jobs:
           distribution: 'temurin'
           java-version: 21
           cache: gradle
+      - name: Move generated sources to correct package
+        run: .\scripts\copyFilesOnBuild.ps1 -inputPath '.\src\main\java\com\microsoft\graph\generated'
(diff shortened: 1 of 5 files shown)


andrueastman:
Sources jar from pipeline artifacts are confirmed to have moved from the generated folder. 
image

javadoc artifacts are now non empty
image

baywet (approved):

